PR #10, opened 2015-03-07
Listen to 'resize' on window, not scrollableParent; fixes #8

### LINKED ISSUE #8: Possible Bug with 'resize' listener
We listen to the 'scroll' and 'resize' events on the 'scrollableParent' element, which is sometimes the window. However, [the 'resize' event is only triggered on the window element][mdn-window-resize], so we may not catch it in cases that the 'scrollableParent' is not the window.

It seems safer to listen to the 'resize' event always on the window, rather than the scrollable parent.

[mdn-window-resize]: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onresize

flarnie:
(From the relevant issue:)
We listen to the 'scroll' and 'resize' events on the 'scrollableParent'
element, which is sometimes the window. However, the 'resize' event is
[only triggered on the window element][mdn-resize-event], so we may not
catch it in cases that the 'scrollableParent' is not the window.

It seems safer to listen to the 'resize' event always on the window,
rather than the scrollable parent.

[mdn-resize-event]: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onresize

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## master (unreleased)
 
+- Fix corner case where scrollable parent is not the window and the window
+  resize should trigger a Waypoint callback.
+
 ## 0.3.0
 
 - Fix Waypoints with the window element as their scrollable parent (Firefox only)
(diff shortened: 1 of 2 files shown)


flarnie:
Should we increment the minor version for this change? I'm not sure it's worth doing so, especially since we want to release 1.0 soon anyway.

lencioni:
I don't think this needs to be in its own release, it can just be part of 1.0. You should add a line to the changelog though.

flarnie:
Sounds good - will add the 'removeEventListener' and the changelog line.

PR #11, opened 2015-03-23
Updated peer dependencies to include React >= 0.12

poteto:


lencioni:
Thanks for the PR! Nothing comes to mind about React 0.13 that would prevent this from working there, but I haven't actually tried it out yet. Have you tried using react-waypoint with React 0.13 yet?

lencioni:
Also, I'm a little bit nervous about opting in to all future versions of React, but maybe it's not a big deal. Can you point to some other projects that do this?

poteto:
I'm in the process of updating our [production app](http://www.homely.com.au) to React `0.13.1`, and so far I haven't encountered any issues with react-waypoint in dev. Would you prefer if I be more explicit about versions, i.e. `~0.13.x`?

lencioni:
> I'm in the process of updating our production app to React 0.13.1, and so far I haven't encountered any issues with react-waypoint in dev.

Awesome! I figured as much but thought it would be best to make sure just in case. In the future, it might be nice to mention details like that in the commit message.

> Would you prefer if I be more explicit about versions, i.e. ~0.13.x?

At least for now, I think that would be best.

poteto:
@lencioni Rebased! Thanks for this component :)

PR #17, opened 2015-06-09
Add event and context

jordansexton:
`_handleScroll` is dispatched by scroll and resize events as well as React lifecycle events. In addition, `onEnter` and `onLeave` should be `call`ed with the context of the parent element being scrolled. These handlers aren't automatically bound to the component with ES6-based React classes, and even for ordinary `React.createClass` components, the handler would be bound to the parent element which is not necessarily the scrollable parent element that Waypoint identifies. It's useful to be able to distinguish between the events and determine the scrolling component, and this change adds no overhead or complexity.

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -83,7 +83,12 @@ const Waypoint = React.createClass({
     return window;
   },
 
-  _handleScroll: function() {
+  /**
+   * @param {Object} event the native scroll event coming from the scrollable
+   *   parent, or resize event coming from the window. Will be undefined if
+   *   called by a React lifecyle method
+   */
+  _handleScroll: function(event) {
     const isVisible = this._isVisible();
 
     if (this._wasVisible === isVisible) {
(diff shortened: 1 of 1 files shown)


trotzig:
Hi @jordansexton, and thanks for contributing to react-waypoint!

The code looks good. I left a few inline comments that you can address, and if I could ask you for one more thing it would be to provide a little more detail in the commit message. For me (and others at Brigade), the most important part of the commit message is to explain the why of the commit. Here's a post that @lencioni wrote on the matter of commit messages: https://medium.com/brigade-engineering/the-secrets-to-great-commit-messages-106fc0a92a25

Again, thanks!

jordansexton:
Revised with jsdoc parameter comment and full commit comment, which I think also answers your previous inline comment on why `this.props.onEnter.call(this, event)` is better than `this.props.onEnter(event)`, but in case it's not clear, here's another explanation.

If `onEnter` (or `onLeave`) is not bound to the parent component passing it in, it will just be a method of `props`, and `this` inside of it will refer to the child `props` object. This is bad because

- it leaks `props` from the event dispatcher to anywhere that scope gets carried to, which is potentially mutable state, but even if not, is an encapsulation violation
- it breaks the expectation that dispatched events are scoped to either a React component or DOM element
- it behaves in an ambiguous or unpredictable fashion with ES5 `React.createClass` vs ES6 `class Component extends React.Component`

If the methods are bound, this does nothing because the bound scope will be used. If not, this behavior is more predictable, conventional, and future-proof.

lencioni:
Thanks for the PR!

PR #18, opened 2015-06-15
Remove call of handleScroll() On componentDidUpdate

viktomas:
- This call was causing multiple onEnter() method calls resulting into
  content being loaded more than on time

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -34,11 +34,6 @@ const Waypoint = React.createClass({
     this._handleScroll();
   },
 
-  componentDidUpdate: function() {
-    // The element may have moved.
-    this._handleScroll();
-  },
-
   componentWillUnmount: function() {
     if (this.scrollableParent) {
       // At the time of unmounting, the scrollable parent might no longer exist.


lencioni:
Thanks for the PR! Can you give me a little more context on why you want to remove this?

IIRC, the reason we added this was to solve the following problem:

1. You render a bunch of content with a `<Waypoint>` at the end of it. This content is tall enough to push the `<Waypoint>` out of the viewport.
2. The content changes, perhaps through some user interaction that removes a bunch of it, which causes the `<Waypoint>` to come into the viewport, but the user hasn't scrolled.

In this event, you would want the `onEnter` function to be called. I'm not convinced that it is a good idea to remove the 'handleScroll()` call from `componentDidUpdate`. Perhaps there is a more elegant way to tackle this--polling comes to mind, but that seems expensive.

viktomas:
Hi, we are never changing content once rendered, so we haven't consider this scenario. The problem we have with this method (and the demo page for waypoint as well) is that the `componentDidUpdate` is called after additional content is loaded. This results every time in loading additional content (can be easily seen on waypoint demo page). We are loading news articles and we need to fetch only one at a time and removing `handleScroll()` fixed that issue.

jeffkole:
@lencioni In the case that you describe, does `componentDidUpdate` even get called?  I would expect that unless the props or state of the `Waypoint` change, `componentDidUpdate` would not be called.  Or, are you suggesting that folks add unused properties to the `Waypoint` instance to trigger updates manually?

trotzig:
I'm closing this since I'm pretty sure the issue has been resolved in https://github.com/brigade/react-waypoint/commit/e3f09f31bb1159e8929829d7db804d8dd250332d

See discussions here
https://github.com/brigade/react-waypoint/issues/35
and here
https://github.com/brigade/react-waypoint/issues/28
for more context. 

I'm sorry this didn't get resolved sooner.

PR #20, opened 2015-07-18
Use "ancestor" instead of "parent"

lencioni:
The term "parent" means that the node is the direct ancestor of the
descendant component (child). This is not necessarily the case in this
code, so I am changing the terminology to use the more general and
accurate "ancestor".

I also made some minor changes that ESLint pointed out.


src/waypoint.jsx
```diff
@@ -19,46 +19,46 @@ const Waypoint = React.createClass({
...
-      // Guarding against this prevents the following error:
+  componentWillUnmount() {
+    if (this.scrollableAncestor) {
+      // At the time of unmounting, the scrollable ancestor might no longer
+      // exist. Guarding against this prevents the following error:
       //
       //   Cannot read property 'removeEventListener' of undefined
-      this.scrollableParent.removeEventListener('scroll', this._handleScroll);
+      this.scrollableAncestor.removeEventListener('scroll', this._handleScroll);
       window.removeEventListener('resize', this._handleScroll);
```
jeffkole: Shouldn't the removal of the `resize` listener from the `window` always occur, regardless of whether or not the `scrollableAncestor` is mounted?

lencioni: Ah yes, I think so. Good call.

PR #33, opened 2015-10-16
Skip `<html>` when looking for scrollable ancestor

zaygraveyard:
The `<html>` tag itself is not scrollable, it causes the window to scroll.

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -323,4 +323,22 @@ describe('<Waypoint>', function() {
       });
     });
   });
+
+  describe('when the <html> is the scrollable parent', () => {
+    beforeEach(() => {
+      // Give the <html> an overflow style
+      document.documentElement.style.overflow = 'auto';
+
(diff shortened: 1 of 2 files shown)


lencioni:
Thanks for the contribution to react-waypoint. I think this looks good and it makes sense to me. However, I'd like to know if you ran into this problem in the wild and what conditions are required to make this be a problem (I have an idea, but I'd like to hear it from you).

Also, will you please add a test to this PR that reproduces this issue? That will help us prevent regressions in the future.

zaygraveyard:
First-off thanks for the awesome component.
I ran into this problem when i was trying to implement an infinite scroll that uses the window as the scrollable container.
And as it happens i had given the `html` element an `overflow: auto` rule to overwrite an earlier `overflow: hidden`, which caused it to be considered as the scrollable parent...
And finally I added a test to check that when the `html` has an `overflow: auto` react-waypoint doesn't throw an error

lencioni:
Thanks!

PR #34, opened 2015-10-21
Add possibility to set className to rendered span

petrbrzek:
Sometimes is necessary to style span because of some non trivial page layout (e.g. using flexbox).

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -19,6 +19,7 @@ const Waypoint = React.createClass({
     // threshold is percentage of the height of the visible part of the
     // scrollable ancestor (e.g. 0.1)
     threshold: PropTypes.number,
+    className: PropTypes.string
   },
 
   /**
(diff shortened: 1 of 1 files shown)


lencioni:
Thanks for your interest in react-waypoint!

Is it possible to solve your issue by wrapping your waypoint with a span and styling that span instead? We believe that passing a className to components is an anti-pattern so we'd prefer to avoid expanding this component's API in this way. More info: https://medium.com/brigade-engineering/don-t-pass-css-classes-between-components-e9f7ab192785

petrbrzek:
Well, it's interesting and also kinda limiting. But I respect it. What about adding some default className  to the `<span>` element like `.waypoint` to be more specific then just the bare element?

trotzig:
I'm closing this without any action. I understand that a class name could be helpful and it isn't hard to add. But we want to stay consistent with how we deal with className props, and this is not the only component where we have been firm on the no-className policy. @petrbrzek I hope you don't feel like we're being overprotecting (I understand it might come across as such). And if adding a wrapper isn't possible in some scenarios, let us know and we can reconsider our policy. :)

Issue #25, opened 2015-09-02
Make Waypoint directional


janpaul123:
E.g. provide methods or configuration so that it not just triggers when Waypoint is in the viewport (as with `onEnter`), but also when it is either above or below the viewport. This would be useful in cases where scrolling happens so fast that `onEnter` is missed (as we saw in our application).

lencioni:
I think this is a good idea. Do you have a proposal for what the API should look like?

janpaul123:
Yeah, not sure. What about an `onAboveViewport` and `onBelowViewport`?

trotzig:
I'm trying to understand what problem you were experiencing. What made scrolling so fast that onEnter/onLeave wasn't fired? Was this on mobile? 

We have to be careful not to confuse people if we add these to the API.  From a glance, it isn't obvious what these two new callbacks mean.

Could we instead fix the "bug" by allowing onEnter and onLeave to be fired if we detect that scrolling was too fast?

janpaul123:
Yeah, that might be a better option. It's impossible to always detect, e.g. if you scroll up and down again, where it was briefly in view but no scroll event was fired. But if it goes from below the viewport to above the viewport then we could detect that.

However, I wonder now if that is what actually happened. Perhaps it's also possible that the Waypoint was instantiated above the viewport in the first place, right above the loading indicators, and that it never moved through the viewport in the first place. I can look into that. In that case fixing the "bug" would not help, but the additional callbacks would.

trotzig:
Yeah, my experience with scroll events is that desktop browsers are pretty good at firing them consistently on every pixel scrolled. I would assume most mobile browsers do too, except for iOS Safari perhaps.

janpaul123:
Oh that's absolutely not the case. This is an example on a random webpage in Chrome:
<img width="544" alt="screen shot 2015-09-03 at 15 55 22" src="https://cloud.githubusercontent.com/assets/177461/9660109/41830b02-5254-11e5-8682-e548536ddab5.png">

trotzig:
Yeah, you're absolutely right. Though I have a hard time seeing the quick up+down scroll not firing events close to the turning point. And even if it can happen (you're scrolling really, really fast), I don't think we can detect it. And I don't think it matters. The scenario where the waypoint was scrolled past too quickly though, we need to detect. I think keeping track of the last seen scrollY and reacting accordingly should be enough, right?

janpaul123:
Yeah, I think so.

[...]

PR #38, opened 2015-10-29
Add direction specific functions

nickclaw:
This adds 4 more optional properties to the Waypoint component: `onEnterFromAbove`, `onEnterFromBelow`, `onLeaveToAbove`, `onLeaveToBelow`.

I haven't updated any tests yet, so don't merge this yet. I just want to make sure this is something that could be accepted before I put too much time into it :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## master (unreleased)
 
+## 1.1.0
+
+- Add second parameter to `onEnter` and `onLeave` callbacks to indicate
+  from which direction the waypoint entered _from_ and _to_ respectively.
+
 ## 1.0.6
 
 - Prevent duplicate onError/onLeave callbacks
--- a/README.md
+++ b/README.md
@@ -58,10 +58,30 @@ var Waypoint = require('react-waypoint');
 
 ```javascript
   propTypes: {
-    onEnter: PropTypes.func, // function called when waypoint enters viewport
-    onLeave: PropTypes.func, // function called when waypoint leaves viewport
-    threshold: PropTypes.number, // threshold is percentage of the height of
-    // the visible part of the scrollable parent (e.g. 0.1)
+
+    /**
+     * Function called when waypoint enters viewport
+     * Both parameters will be null if the waypoint is in the
+     * viewport on initial mount.
+     *
+     * @param {Event|null} event
+     * @param {'above'|'below'|null} from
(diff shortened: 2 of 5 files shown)


nickclaw:
Could close #25

trotzig:
Hi @nickclaw, and thanks for contributing to react-waypoint!

This looks useful, but it would be good to understand the usecase a little better. 

Instead of adding the new props, what do you think about passing the previous and current position to the callbacks? We would have to combine it with the native event thats being passed in, but how about something like this

onLeave({ previousPosition: 'inside', currentPosition: 'below', nativeEvent: event});

lencioni:
Another option would be to pass direction or edge to the `onEnter` and `onLeave` callbacks. Something like:

```js
onLeave(event, 'bottom')
```

nickclaw:
@trotzig My use case is something like Apple Calendar, where I have a Waypoint component next to each month, when the Waypoint leaves from the top I want to change the fixed title, and when it enters from the top I want to revert the title to the previous month.

screen



@lencioni that was what I initially tried, but I didn't like the inconsistency between the meaning of the second parameter between the two functions.

```
onLeave(event, currentLocation)
onEnter(event, previousLocation)
```

I think my solution is easier on users because they can just declare what specific events they want to be notified about, instead of having to compare string constants to figure out what's going on. But I can see why adding a property for every possibility isn't as scalable if this module ever handles horizontal scrolling.

What do you guys think? I wouldn't mind changing my pull request to match @lencioni's solution :)

trotzig:
Thanks for the use case description and the gif (that looks great by the way!), it helps us to understand the problem better. I like @lencioni's suggestion in that it doesn't break backwards compatibility with the native `event` being passed down. 

I think it's okay for consumers to have to compare strings for this type of use case. So I'd say go with what @lencioni said. It would be great if you could add a test for this. And make sure to add your use case to the commit message, some documentation to README.md, plus a short description in CHANGELOG.md.

nickclaw:
Alright feeling pretty good about the state of the pull request. I didn't update `package.json`, but I figure this is a minor version bump.

trotzig:
Thanks @nickclaw!

also: lol, I just realized that gif is from the actual Apple Calendar... It would be fun to see your implementation though!

trotzig:
I'll cut a new release in a few hours, got to run out now.

PR #40, opened 2015-11-02
Ensure component is mounted when executing scroll handler

lencioni:
In the Brigade app, we've seen an error occasionally happen in
production where `this === null` inside of the `_currentPosition`
function. This causes `ReactDOM.findDOMNode` to error out.

While I'm not entirely sure I understand how this could happen, my
hypothesis is that the component has become unmounted before the scroll
handler can be executed, causing a race condition. Ensuring that the
component is mounted before executing the handler should prevent this
from happening.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## master (unreleased)
 
 - Add statics for edge argument used by `onEnter` and `onLeave`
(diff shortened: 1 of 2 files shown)


janpaul123:
Hm, yeah, this baffles me too. Feel free to merge.

trotzig:
Already done!

PR #42, opened 2015-11-05
Allow ancestor to be enforced to be window. Check for window in lifec…

khankuan:
…ycle methods for server-side rendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,17 @@ var Waypoint = require('react-waypoint');
      * Threshold - a percentage of the height of the visible
      * part of the scrollable parent (e.g. 0.1)
      */
-    threshold: PropTypes.number
+    threshold: PropTypes.number,
+
+    /**
+     * Scrollable Parent - A custom parent to determine if the
+     * target is visible in it. This is useful in cases where
+     * you do not want the immediate scrollable ancestor to be
+     * the container. For example, when your target is in a div
+     * that has overflow auto but you are detecting onEnter based
+     * on the window.
+     */
+    scrollableParent: PropTypes.any
   },
 ```
 
--- a/build/npm/waypoint.js
+++ b/build/npm/waypoint.js
@@ -22,7 +22,13 @@ var Waypoint = React.createClass({
     // scrollable ancestor (e.g. 0.1)
     threshold: PropTypes.number,
     onEnter: PropTypes.func,
-    onLeave: PropTypes.func
+    onLeave: PropTypes.func,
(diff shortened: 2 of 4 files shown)


trotzig:
It looks like this change has two purposes:

1. Making sure that waypoints can be rendered in environments without a `window` global (i.e. on the server)
2. Allowing the `scrollableParent` to be overridden

For 1, I'd be happier with an approach that just short-circuited (early return) in a few places:

```jsx
componentDidMount() {
  if (!window) { return; }
  ...
}

componentDidUpdate() {
  if (!window) { return; }
}
```

For 2, did you consider making it more general by allowing a `scrollableParent` element to be passed in? That would make it more useful than the `window` boolean. 

For both these changes, it would be good if the commit messages explained the usecase a bit more. Also, make sure to use a blank line between the commit subject and the message, e.g.

```
Allow react-waypoint to be used on the server

We have an isomorphic React application that renders pages on the server before sending them down to the client 

[... and so on...]

```

Tests would also be useful, especially since these things aren't widely used (and therefore more likely to be overlooked when code is changed/refactored).

khankuan:
Good pointers! Would look at them at free time ;)

trotzig:
Thanks for updating @khankuan! I know it's frustrating to hear us bounce back with change requests, but I'm going to ask you to do a few things before we can merge this. :tennis: 

First of all, the two changes in this PR should be broken up into two commits (and perhaps even two PRs). Right now, I'd be sort of happy to merge the server-side support (short-circuit lifecycle methods if window is undefined). We do need a test for it first though as this is the sort of thing that's easy to break.

In order to merge the `scrollableParent` addition we need you to explain the change better, preferably in the commit message. Did you run into an issue that brought you down that path or is it a nice-to-have? A minor thing about that too is that you should update the README file with an explanation of it. Don't worry about getting the text "right", we can always adjust it after merge if we think it's important.

Again, thanks for contributing.

khankuan:
How would you like to approach server-side testing?

trotzig:
Good question. If we can somehow make `window` undefined in a test, we can verify that the component can render. In order to do that, we might have to do some dependency injection. One idea would be to have a static method in the component that returns `window` that we can then mock in tests (in Jasmine, that would mean something like `spyOn(Waypoint, 'window').and.returnValue(undefined)`.

trotzig:
Sorry @khankuan for dropping the ball on this. I actually missed that you had updated this PR. I'll merge this right away!

PR #45, opened 2015-11-18
fix document and html detection. Previous solution does not work well with iframe

petrbrzek:


lencioni:
Thanks for the pull request! Would you please add a test that reproduces the issue to prevent regressions?

petrbrzek:
Sorry it was probably false positive. I've tested previous solution in my app today and no bug happened. So it's probably safe to close it.

lencioni:
Alright, thanks for looping back!

petrbrzek:
Ok, it is a real problem but I'm unable to reproduce it every time (the pull request fixes it because of better document detection). Here is the screenshot. 

avocode_-_2_4_5__dev_mode_

petrbrzek:
We should probably reopen this one @lencioni.

Issue #44, opened 2015-11-13
Remove bower support


lencioni:
We haven't updated our bower.json file since 1.0.2, and it seems that Bower may be dying. Let's clean it out.

lencioni:
I believe that this will allow us to remove the checked in build/npm directory and just let npm take care of that like we do in react-simple-pie-chart.

PR #46, opened 2015-11-18
Remove bower support

lencioni:
We haven't updated our bower.json file since 1.0.2, and it seems that
Bower may be dying. Let's clean it out.

Finishes #44.


lencioni:
Bump!

trotzig:
I'll rebase and merge this tomorrow if no one else gets to it before.

PR #48, opened 2015-12-12
Calculate proper offset when <html> or <body> has a margin

janpaul123:
Because offsetTop doesn't take that into account. Instead I used node.getBoundingClientRect with some adjustments, based on jQuery:

https://github.com/jquery/jquery/blob/29370190605ed5ddf5d0371c6ad886a4a4b5e0f9/src/offset.js#L110

This also has the advantage that if window is the scroll container, we get the result immediately (instead of using recursion) which might be slightly faster.

### DIFF
--- a/build/npm/waypoint.js
+++ b/build/npm/waypoint.js
@@ -147,6 +147,11 @@ var Waypoint = React.createClass({
       throw new Error('The scrollable ancestor of Waypoint needs to have positioning to ' + 'properly determine position of Waypoint (e.g. `position: relative;`)');
     }
 
+    if (this.scrollableAncestor === window) {
+      var rect = node.getBoundingClientRect();
+      return rect.top + window.pageYOffset - document.documentElement.clientTop;
+    }
+
     if (node.offsetParent === this.scrollableAncestor || !node.offsetParent) {
       return node.offsetTop;
     } else {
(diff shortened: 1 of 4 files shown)


trotzig:
Thanks for finding and fixing! I know that `offsetTop` and `getBoundingClientRect` both suffer from performance issues. I tried to look at some jsPerf benchmarks but for some reason the graphs don't show up on any jsPerf examples for me at the moment.

https://jsperf.com/offset-vs-getboundingclientrect/4

janpaul123:
Ah, you're right — at least on my machine (Chrome 47.0.2526.80 on Mac OS X 10.11.0) getBoundingClientRect is 85% slower. However, I don't see a more reliable way of calculating this that takes into account all borders/paddings/margins of <html> and <body> — we can do all that manually, but my commit seems more robust. What do you think?

trotzig:
I think it's fine to merge this as-is, and think about solutions if we notice that this is real problem. Then we might consider using some kind of heuristic to figure out if we need to do the expensive calculation or not. I don't know if such exists though. 

Another option would be to list this issue in the "Limitations" section in the Readme. I would think that 99% of all websites have a margin-less body. (Actually, why did we end up with a margin?).

janpaul123:
It's because of the branch banner, that interacts with the `position:relative` that we have on <body>. It seems relatively hard to refactor that out, and I rather have <Waypoint> work in more cases. Merging this and cutting a release!

PR #51, opened 2015-12-30
Update some things

lencioni:


lencioni:
FYI, this PR is very similar to [the one I made over here](https://github.com/brigade/react-simple-pie-chart/pull/7), so it might be worth reviewing them at the same time.

trotzig:
Looks good, thanks @lencioni!

trotzig:
Doh, I meant to merge the other PR (the one you reference above...). Oh well, doing some post-merge review now instead. :)

PR #54, opened 2016-01-19
New prop added - fireOnRapidScroll

lordakshaya:
The new Prop is to allow the developers to disable firing of onEnter/onLeave events on rapid scrolling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 node_modules
 npm-debug.log
+.idea/
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ var Waypoint = require('react-waypoint');
   onEnter={this._handleWaypointEnter}
   onLeave={this._handleWaypointLeave}
   threshold={0.2}
+  fireOnRapidScroll=true
 />
 ```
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 build
 node_modules
 npm-debug.log
+.idea/
```
trotzig: (Shameless plug for a tool I built): Using WebStorm? Care to develop a plugin for import-js, a tool I've built to help importing/requiring javascript dependencies? There are plugins for Emacs, vim, and Sublime. But none for WebStorm (yet). https://github.com/trotzig/import-js/issues/102

lordakshaya: @trotzig could do that, let me give it a look. Most prolly, I'll be needing info/guidance on how to do it. :)

lordakshaya:
@trotzig cool, let me know.

.gitignore
trotzig: I added some more information to the import-js issue: https://github.com/trotzig/import-js/issues/102#issuecomment-173168629
And I'm happy to help with whatever you need.

trotzig:
I thought I posted a general comment here too, but I must have forgot to actually hit the submit button... Here goes:

First of all, thanks for contributing! I can see this being useful in a couple of use cases, and it would be good if you could briefly describe the one you had in mind. What's missing from this commit are some tests to verify the new behavior. I'll go ahead and do that in a follow-up commit, I'll make sure to tag you so that you can make sure that I got it right.

PR #57, opened 2016-01-27
Operate with body.offsetParent == null

just-paja:
I was getting issues when the Waypoint was placed in a scrollable div and I tracked it down to this exception: "The scrollable ancestor of Waypoint needs to have positioning ...", but the positioning was there. It was failing because document.body.offsetParent is always null.

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -335,16 +335,6 @@ describe('<Waypoint>', function() {
     });
   });
 
-  describe('when the scrollable parent does not have positioning', () => {
-    beforeEach(() => {
-      delete this.parentStyle.position;
-    });
-
-    it('throws an error', () => {
-      expect(this.subject).toThrow();
-    });
-  });
-
   describe('when the window is the scrollable parent', () => {
     beforeEach(() => {
       // Make the normal parent non-scrollable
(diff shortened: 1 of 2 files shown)


trotzig:
Thanks for contributing to react-waypoint @just-paja!

Just to better understand the issue you are facing: is it happening because you have an `overflow` or an `overflow-y` set on the `<body>` element? I agree that the error is confusing, though I wonder if instead we could change the `_findScrollableAncestor` slightly so that we don't use body as the scrollable parent (and instead just fall back to `window`)?

It would be good to have a regression test for this too so that we don't accidentally break it at some later point.

BizzlePop34:
This change fixed the problem I was having.  I am still looking for the root cause of the issue.

just-paja:
@trotzig, `overflow-y` was given to element within the document and I noticed that `_distanceToTopOfScrollableAncestor` is going trough chain of offsetParent(s) until it reaches the body. Give me please some time to put in real example.

```html
<html>
<body>
  <div name="irrelevant div">
    <header></header>
    <div style="position:absolute; left:0; top:150px; right:0; bottom:0; overflow-y:scroll;">
      <div name="some container">
        <Waypoint />
        <div style="height: 1000px"></div>
        <Waypoint />
      </div>
    </div>
  </div>
</body>
</html>
```

trotzig:
This was fixed (I believe) in a6dc4d9ab8a560fc. Let me know if that's not the case and we can reopen this PR.

PR #59, opened 2016-01-29
scrollableParent prop type fix

alex-ionochkin:
When `scrollableParent` prop is set to a DOM element, waypoint writes a warning `Warning: Failed propType: Invalid prop scrollableParent supplied to Waypoint, expected a ReactNode.` to the browser console
This is a fix for that warning according to what README says about `scrollableParent` prop type

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -14,7 +14,7 @@ const propTypes = {
   onEnter: PropTypes.func,
   onLeave: PropTypes.func,
   fireOnRapidScroll: PropTypes.bool,
-  scrollableParent: PropTypes.node,
+  scrollableParent: PropTypes.any,
 };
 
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks for taking the time to prepare a PR @alex-ionochkin!

There's probably a better proptype than `any` here, but I'm not sure which one it is. I'm looking at the [`PropTypes documentation`](https://facebook.github.io/react/docs/reusable-components.html) and there's no type for DOM elements. Which makes sense because A) you should be using React components, and B) in order to gain access to a DOM element in a React tree you need to render it at least once first. However I think the most common thing to pass in as `scrollableParent` is `window` or perhaps `document.body`.

We could perhaps get away with using `PropTypes.object`, but I think what you have in this PR is fine.

PR #65, opened 2016-03-16
Converts ancestor positioning error to warning

### LINKED ISSUE #64: Waypoint scrollable ancestor errors should really be warnings
I have a use case in which I render a waypoint inside of a `display: none` ancestor (which is then displayed shortly after rendering). After converting the errors to warnings (`throw new Error` --> `console.warn`), everything continued to work as expected.

kenaniah:
Fixes #64

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -155,7 +155,7 @@ export default class Waypoint extends React.Component {
    */
   _distanceToTopOfScrollableAncestor(node) {
     if (this.scrollableAncestor !== window && !node.offsetParent) {
-      throw new Error(
+      console.warn(
         'The scrollable ancestor of Waypoint needs to have positioning to ' +
         'properly determine position of Waypoint (e.g. `position: relative;`)'
       );


trotzig:
Interesting. I hadn't thought of the fact that you might render the waypoint in a hidden container. I wonder if changing to a warning gets to the root of the problem though. Can we detect this some other way, and avoid the warning (and error) altogether? There's also a workaround you can apply in your application code: avoid rendering the waypoint until the container is visible. That may or may not be easy, depending on how your element hierarchy looks like.

Any direction you take, we'll need the tests fixed to reflect the new behavior.

Thanks for taking the time to put together a PR @kenaniah!

tomdcc:
We are seeing something similar - when our component is mounted, for some reason `node.offsetParent` is always null - it only becomes non-null later on. So we get the same error.

Changing this to a warning would at least let us proceed, though it's not ideal. Simply not triggering any events until `node.offsetParent` is there is one option. Another would be to keep the warning (as it probably often helpful) but pass in some sort of flag that suppresses the warning for situations where this behaviour is needed.

trotzig:
Thanks for providing more context @tomdcc. If you can figure out why `node.offsetParent` is null, drop a line in this thread. 

The main issue I have with simply changing to a warning is that it will cause the `_distanceToTopOfScrollableAncestor(node)` to return something that we then use to calculate whether or not the waypoint is in view. If this number is incorrect, the waypoint won't work the way it's supposed to. I wonder if we should introduce a new position called `invisible`. We can throw a specific error from `_distanceToTopOfScrollableAncestor(node)` to signal this state, and catch that error in `_currentPosition`. Then we won't have to throw any error, nor warn about it. And people can inspect the waypoint component using React devtools to figure out that it ends up in this state. 

Since it's probably easier for you to try this out @kenaniah and @tomdcc, would any of you like to take it on? If not I'll get to it when I have a few cycles over, but I can't promise that will be any time soon.

tomdcc:
@trotzig I've submitted PR #67 which introduces an `invisible` position.

trotzig:
Closing this PR since it was fixed by #67. Thanks everyone!

PR #67, opened 2016-03-23
Handle waypoint parent not being visible

tomdcc:
This can happen for a variety of reasons, e.g. a transition animation.

Rather than throwing an exception, we set the position to the new "invisible"
position. When the component becomes visible and we notice in the next scroll
event (or forced update), and standard event handlers will fire as we notice
the change from "invisible" to whatever the new state is.

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -338,20 +338,24 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
     beforeEach(() => {
       delete this.parentStyle.position;
-      this.props.onEnter.calls.reset();
-      this.props.onLeave.calls.reset();
     });
 
-    it('does not call handlers to begin with', () => {
-      expect(this.subject).not.toThrow();
+    it('does not call handlers until node becomes visible', () => {
+      // initial render hidden
+      this.component = this.subject();
       expect(this.props.onEnter).not.toHaveBeenCalled();
       expect(this.props.onLeave).not.toHaveBeenCalled();
-    });
+      const node = ReactDOM.findDOMNode(this.component);
 
-    it('does call handlers when node becomes visible', () => {
-      this.parentStyle.position = 'relative';
-      expect(this.subject).not.toThrow();
+      // now show it and we should get an onEnter
+      node.style.position = 'relative';
+      scrollNodeTo(this.component, 10);
       expect(this.props.onEnter).toHaveBeenCalled();
+
+      // now hide and we should see an onLeave
+      node.style.display = 'none';
+      scrollNodeTo(this.component, 20);
+      expect(this.props.onLeave).toHaveBeenCalled();
     });
   });
 
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -338,10 +338,20 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
     beforeEach(() => {
       delete this.parentStyle.position;
+      this.props.onEnter.calls.reset();
+      this.props.onLeave.calls.reset();
     });
 
-    it('throws an error', () => {
-      expect(this.subject).toThrow();
+    it('does not call handlers to begin with', () => {
+      expect(this.subject).not.toThrow();
+      expect(this.props.onEnter).not.toHaveBeenCalled();
+      expect(this.props.onLeave).not.toHaveBeenCalled();
+    });
+
+    it('does call handlers when node becomes visible', () => {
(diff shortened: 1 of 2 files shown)


trotzig:
Thanks @tomdcc! I'll have a closer look at this later tonight.

src/waypoint.jsx
```diff
@@ -169,8 +167,8 @@ export default class Waypoint extends React.Component {
     if (node.offsetParent === this.scrollableAncestor || !node.offsetParent) {
       return node.offsetTop;
     } else {
-      return node.offsetTop +
-        this._distanceToTopOfScrollableAncestor(node.offsetParent);
+      let nextOffset = this._distanceToTopOfScrollableAncestor(node.offsetParent);
```
trotzig: The build is failing because you don't need `let` here. We lint the code (eslint) as part of testing, and this error comes up:
```
 170:11  error    `nextOffset` is never modified, use `const` instead  prefer-const
```

spec/waypoint_spec.js
```diff
@@ -338,10 +338,20 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
     beforeEach(() => {
       delete this.parentStyle.position;
+      this.props.onEnter.calls.reset();
+      this.props.onLeave.calls.reset();
```
trotzig: You don't need to reset these.

spec/waypoint_spec.js
```diff
@@ -338,10 +338,20 @@ describe('<Waypoint>', function() {
...
+      expect(this.props.onLeave).not.toHaveBeenCalled();
     });
 
-    it('throws an error', () => {
-      expect(this.subject).toThrow();
+    it('does call handlers when node becomes visible', () => {
+      this.parentStyle.position = 'relative';
+      expect(this.subject).not.toThrow();
+      expect(this.props.onEnter).toHaveBeenCalled();
     });
```
trotzig: In order for this to test the right thing, you need to do something like this:

```js
    it('does call handlers when node becomes visible', () => {
      this.component = this.subject();
      ReactDOM.findDOMNode(this.component).style.position = 'relative';
      scrollNodeTo(this.component, 10);
      expect(this.props.onEnter).toHaveBeenCalled();
    });
```

`this.subject()` will make the component mount on the page. The `this.parentStyle.position` only applies to the first render (inside `this.subject()`). So what you're really testing here isn't a transition from invisible to visible, it's just making sure that `onEnter` is called on initial render. My modification means we alter the DOM node after it has been rendered (which is what we want), and then the scroll makes sure that we recalculate positioning.

spec/waypoint_spec.js
```diff
@@ -338,10 +338,20 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
     beforeEach(() => {
       delete this.parentStyle.position;
+      this.props.onEnter.calls.reset();
+      this.props.onLeave.calls.reset();
+    });
+
+    it('does not call handlers to begin with', () => {
+      expect(this.subject).not.toThrow();
```
trotzig: You could change this to `this.subject();`, since the throw is a thing of the past.

trotzig:
This is looking good, thanks for taking the time! I left a few comments on how I think you need to change the code around to make the test succeed and test the right thing.

src/waypoint.jsx
tomdcc: Weird that that wasn't picked up... ah, when you run the test script the eslint is run once initially and then karma runs continuously - I hacked away until the tests were green and I was happy with the code, but never ran the tests from scratch after that.

spec/waypoint_spec.js
tomdcc: Thanks for explaining. I understood what was going on but wasn't sure how to accomplish what I wanted, which is more or less what you've shown above.

spec/waypoint_spec.js
tomdcc: I ended up merging these tests into a single test

tomdcc:
Pushed update after your feedback, thanks for the fast turnaround!

src/waypoint.jsx
trotzig: Ah, I see how that could happen. :)

spec/waypoint_spec.js
trotzig: I was actually going to suggest that. Good call. :)

trotzig:
Thanks @tomdcc! I'll wait for your feedback on https://github.com/brigade/react-waypoint/pull/68 before carving a new release.

tomdcc:
Thanks!

PR #68, opened 2016-03-24
Add onPositionChange event

tomdcc:
While I'm doing some stuff in this codebase, here's another thing that would make life a bit easier for us.

When a waypoint's position has changed, we often just want to know what the new position is. Currently we have to write both `onEnter` and `onLeave` handlers to convert this info back into a single position update, but since the Waypoint internally already has this as a single position change, it seems like this is an easy and useful thing to expose.

While the existing events pass (event, new/old position) as the params, I consciously decided not to follow the convention and passed (new position, old position, event) in that order, since the most frequent use case is just wanting to know the new position, so it's a pain to have to write handlers that take an event as the first param only to throw it away.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,17 @@ can then use a `key` prop to control when a waypoint is reused vs. re-created.
 />
 ```
 
+Alternatively, you can also use an `onPositionChange` event to just get
+notified when the waypoint's position (e.g. inside the viewport, above or
+below) has changed.
+
+```javascript
+<Waypoint
+  onPositionChange={this._handlePositionChange}
+/>
+```
+
+
 ### Example: [JSFiddle Example][jsfiddle-example]
 
 [jsfiddle-example]: http://jsfiddle.net/L4z5wcx0/7/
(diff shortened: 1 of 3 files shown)


trotzig:
Looks good overall. I agree about the order of arguments to onEnter/onLeave being weird. We should fix that, so [I added an issue to track it](https://github.com/brigade/react-waypoint/issues/69).

I'm wondering if we should start passing arguments to callbacks in an object instead. By looking at your code, I wasn't sure at first what order the positions came in. So I'm suggesting bundling them all in an object, so that you can do this:

```jsx
<Waypoint onPositionChange={({ before, after, event }) => { ... }); /> 
```
...if you're using es6 destructuring, or
```jsx
<Waypoint onPositionChange={(props) => { console.log(props.before, props.after, props.event) }); /> 
```

Before we merge this, can you please add something to the readme?

tomdcc:
Object is probably the best approach - destructuring makes it work pretty nicely for only the info that you're interested in, e.g. :

```jsx

handlePositionChange({newPosition}) {
    // etc
}}}

<Waypoint onPositionChange={handlePositionChange}/>
```

I personally can be :+1: for that since we're using ES6 here, but whether to impose that on your other users is your call.

Anyway it's a breaking change, so I'll leave it up to you to decide when is best to do it.

tomdcc:
Updated the README

trotzig:
I think we should 
1. Introduce it for `onPositionChange` (which is new so no breaking change)
2. Release 1.4.0 with that change
3. Update `onEnter` and `onLeave` to follow the same structure. 
4. Release 2.0.0 with the breaking change

How does that sound? If you take care of 1 (in this PR), I'll do 2-4.

trotzig:
I decided to release 1.3.1 with the fix for invisible parents before we merge any of this, I hope that was okay. Did you see my last comment about the proposed changes? If you don't feel like making those changes or don't have time, I can pick it up from here. :)

trotzig:
No worries, I can take it from here.

trotzig:
I added a commit on top of yours and merged to master. I'll release a new version in the coming days. Thanks for contributing @tomdcc!

tomdcc:
Awesome! Thanks!

PR #71, opened 2016-04-05
not working from other "page"

Kepro:
the problem is that if we start page with waypoint then we get correct element from _findScrollableAncestor
eg. some <div> but if we come from other page and come back then we get Window element

i test it just with 
````
        this.scrollableAncestor = this._findScrollableAncestor();
        console.log('this.scrollableAncestor', this.scrollableAncestor);
````

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -43,11 +43,13 @@ export default class Waypoint extends React.Component {
       return;
     }
 
-    this._handleScroll = this._handleScroll.bind(this);
-    this.scrollableAncestor = this._findScrollableAncestor();
-    this.scrollableAncestor.addEventListener('scroll', this._handleScroll);
-    window.addEventListener('resize', this._handleScroll);
-    this._handleScroll(null);
+    setTimeout(() => {
+      this._handleScroll = this._handleScroll.bind(this);
+      this.scrollableAncestor = this._findScrollableAncestor();
+      this.scrollableAncestor.addEventListener('scroll', this._handleScroll);
+      window.addEventListener('resize', this._handleScroll);
+      this._handleScroll(null);
+    });
   }
 
   componentDidUpdate() {


trotzig:
Hi @kepro, thanks for taking the time to dive into the code! I think we need a little more context in order to get to the root of the problem however. What do you mean by "other page"? Are you doing client-side routing (something like [react-router](https://github.com/reactjs/react-router)) or are you using full page loads/server-side routing? 

It's interesting to see that adding a defer (`setTimeout`) fixes the problem. Are you applying css/styling asynchronously in some way? Or perhaps a transition when the page changes?

Kepro:
Hi @trotzig !

thanks for your fast response
sorry, yes i'm using react-router 
so i have /page and /page2 on /page is waypoint
if i start browser with /page then waypoint is working then i go to /page2 and back to /page then not working

also if i start browser with /page2 and go to /page then is not working

due if i come back from /page2 i get _findScrollableAncestor Window and not my Div 

also this bug can be reproduce in firefox and chrome

trotzig:
Can you share a little bit of your code? If possible reduce it into a small reusable case. 

Some questions I can think of directly:
- is the waypoint rendered inside a "page" component, or completely outside? 
- are you using animation when you transition from one page to another?

trotzig:
Any progress here @Kepro? It would be good to further understand why you are experiencing issues. It sounds like there is a bug somewhere (though I'm not sure where).

Kepro:
For questions is in page > lef panel > waypoint
And No without transition 

I rly want to create a simple demo it im busy now,  but maybe tomorrow

Kepro:
@trotzig, so you try with react router or not?

trotzig:
Yes, we use react-router on brigade.com, and waypoints fire as they should.

Kepro:
and way point is attachet to DIV or to window?

trotzig:
I think most of our waypoints use the window as the `scrollableAncestor`.

Kepro:
so that is the main problem, try waypoint just in scrollable div on my side is in left panel and then you get the same problem as i can reproduce

trotzig:
Do you think you can come up with a minimal repro case? Or at least try removing things one by one and see if the problem at some point goes away? I don't feel like we have reached the root of the problem yet.

Also, a screenshot of how the page is structured might help in understanding what's going on.

Kepro:
so here is my structure and console
"start page"
http://prntscr.com/ar8xnu
"come to page" 
http://prntscr.com/ar8xps

i think that componentDiDmount is fired faster than browser render the page and that is why setTimeout with zero time is working

my react version is 0.14.8
and waypoint is 1.3.1

Kepro:
omg... i think that i found bug in 3th party css loading... if i come to page css is not loaded(rendered) so waypoint doesn't see scrollable group just after timeout... so the problem is i think with decorator "withStyles" sorry for spend your time here :/

trotzig:
That's okay! I'm glad you got it figured out.

PR #72, opened 2016-04-07
Add displayName for easier debugging

Mordeaux:
Setting `displayName` ensures that developers can search for <Waypoint> components more easily in the Chrome Developer Tools React extension.


trotzig:
Thanks!

Issue #74, opened 2016-04-08
React 15 support


axelg12:
With the newly released React 15, the peer dependency is outdated and will not install.
```
react-waypoint@2.0.0 requires a peer of react@^0.14.0 but none was installed.
react-waypoint@2.0.0 requires a peer of react-dom@^0.14.0 but none was installed.
```
The quick way to fix this, I think, is to change the line from 
```
"peerDependencies": {
    "react": "^0.14.0",
    "react-dom": "^0.14.0"
  },
```
to 
```
"peerDependencies": {
    "react": "^0.14.0 || ^15.0.0",
    "react-dom": "^0.14.0 || ^15.0.0""
  },
```
This will allow people to use it with both 0.14 and 15.0

trotzig:
That sounds good to me. Would you mind putting that together in a PR? You should bump the devDependency at the same, so that we start developing (and testing) with 15.0, but allow peer dependencies back to 0.14.

trotzig:
Whoops, I didn't realize someone else beat us. :)

trotzig:
Fixed by a6b86453d5.

PR #75, opened 2016-04-09
Add React 15.0 support

kamui:
Related to: https://github.com/brigade/react-waypoint/issues/74

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   "author": "Brigade Engineering",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^0.14.0",
-    "react-dom": "^0.14.0"
+    "react": "^0.14.0 || ^15.0.0",
+    "react-dom": "^0.14.0 || ^15.0.0"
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks! Would you mind bumping the dev dependencies to 15.0 as well? See #74 for some more context.

kamui:
Oops, updated that too.

trotzig:
Sorry for the back-and-forth, but could you change so that the devDependency is ^15.0.0 only? That would make dev and our tests more predictable since we're moving away from 0.14 anyway. There is of course a risk that we add funcionality that's not compatible with 0.14, but let's worry about that then.

kamui:
No problem, the dev deps only are now ^15.

trotzig:
Thanks for the quick turnaround! I'll merge this and push a new version later tonight (in a few hours).

kamui:
Great, thanks! This was the last module in my app that was locked to 0.14.

PR #76, opened 2016-04-11
Update changelog

lencioni:
Some of the commit messages in the git log were a bit hard to follow, so
I'm not sure I got it all in correctly, but it is definitely better than
what was here before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,36 @@
 ## master (unreleased)
+
+
+
+## 2.0.1
+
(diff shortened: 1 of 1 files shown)


trotzig:
I think you got it all. I actually forgot that we kept this log.

lencioni:
Thanks for merging! This is really helpful for people who want to upgrade without breaking things.

trotzig:
Cool. Let me know if you run into any issues after upgrading!

PR #79, opened 2016-04-27
Add babel-plugin-add-module-exports

kmayer:
It's in your .babelrc, but our project is using Babel v6, and hasn't needed it yet. 

Until I added this to *our* devDependencies, we got this error:

```
11:33:53 webpack.1 | ERROR in /Users/pacerpro/workspace/pacerpro/~/react-waypoint/build/npm/waypoint.js
11:33:53 webpack.1 | Module build failed: ReferenceError: Unknown plugin "add-module-exports" specified in "/Users/pacerpro/workspace/pacerpro/node_modules/react-waypoint/.babelrc" at 1, attempted to resolve relative to "/Users/pacerpro/workspace/pacerpro/node_modules/react-waypoint"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
   "license": "MIT",
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0"
+    "react-dom": "^0.14.0 || ^15.0.0",
+    "babel-plugin-add-module-exports": "^0.1.2"
   },
   "devDependencies": {
     "babel": "^6.0.0",


lencioni:
I'm a little confused by this. I don't think we want to add this plugin to peerDependencies. It seems that maybe babel-loader is trying to run on things in node_modules, which you shouldn't need. Can you put together a small and self contained repro case?

kmayer:
I think I jumped the gun on this, just close it.

lencioni:
In case someone ends up here with the same problem, can you share what you did to resolve this?

kmayer:
It might have to do with my webpack config. When we pushed to CI, the build didn't break, which leads me to believe that it was probably something stuck on my local box.

I just added the plugin to my own top-level dev dependencies in package.json and re-ran npm install.

PR #81, opened 2016-04-28
Refactor _currentPosition to use getBoundingClientRect()

kmayer:
This is a pretty clean refactor. Only 1 test had to be modified (that one does bother me, though. See my note, below).

Here's why we did the refactor...

We have a component, `<StickyHeadingList>`, (https://gist.github.com/kmayer/8f1ea347cf688b0bc0f866856ec9aedc) which uses the following format:

```html
<StickyHeadingList>
  <section class="sticky-section">
    <header>Sticky-ish header</header>
    <div>content</div>
    <div>content</div>
    ...
  </section>
  <section class="sticky-section">
    <header>Next sticky-ish header</header>
    <div>more content</div>
    <div>more content</div>
    ...
  </section>
</StickyHeadingList>
```

in the component, this gets mapped to:

```jsx
<div className="sticky-heading-list">
  <article ref="stickyList" className="sticky-list-wrapper" onScroll={this.handleScroll}>
    {this.props.children}
  </article>
</div>
```

with the following style sheet:

```scss
.sticky-heading-list {
  height: 100%;
  overflow: hidden;
  position: relative;

  // The main list
  & article.sticky-list-wrapper {
    height: 100%;
    overflow-x: hidden;
    overflow-y: scroll;

    // Section headers when "sticky"
    & section.sticky-section.sticky header {
      position: absolute;
      top: 0;
    }
  }
}
```

The problem is that when `<Waypoint>` calculates `_distanceToTopOfScrollableAncestor`, it uses `offsetParent`, which skips over the scrolling element `<article.sticky-list-wrapper>` in favor of `<div.sticky-heading-list>` (because it is positioned) which doesn't scroll and is 100% of the view port. So the return value is always `null`. 

By replacing the algorithm, it works independent of offsetParent, does not need to recurse up the DOM tree and eliminates a private method.

I'm concerned that the test on line 529, "does not call handlers until node becomes visible," does not actually test the right thing any more. I don't understand how deleting the parents position, then making it `relative` again makes it "visible". I admit that it's my own lack of understanding.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,6 +6,5 @@
 CHANGELOG.md
 karma.conf.js
 spec
-src
 tests.webpack.js
 webpack.conf.js
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-waypoint",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "description": "A React component to execute a function whenever you scroll to an element.",
   "main": "build/npm/waypoint.js",
   "repository": {
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -25,9 +25,9 @@ const scrollNodeTo = function(node, scrollTop) {
 describe('<Waypoint>', function() {
   beforeEach(() => {
     this.props = {
-      onEnter: jasmine.createSpy(),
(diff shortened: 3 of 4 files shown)


kmayer:
I'm going to move this PR off of my master and onto a clean branch.

PR #82, opened 2016-04-28
Refactor _currentPosition to use getBoundingClientRect()

kmayer:
This is a pretty clean refactor. Only 1 test had to be modified (that one does bother me, though. See my note, below).

Here's why we did the refactor...

We have a component, `<StickyHeadingList>`, (https://gist.github.com/kmayer/8f1ea347cf688b0bc0f866856ec9aedc) which uses the following format:

```html
<StickyHeadingList>
  <section class="sticky-section">
    <header>Sticky-ish header</header>
    <div>content</div>
    <div>content</div>
    ...
  </section>
  <section class="sticky-section">
    <header>Next sticky-ish header</header>
    <div>more content</div>
    <div>more content</div>
    ...
  </section>
</StickyHeadingList>
```

in the component, this gets mapped to:

```jsx
<div className="sticky-heading-list">
  <article ref="stickyList" className="sticky-list-wrapper" onScroll={this.handleScroll}>
    {this.props.children}
  </article>
</div>
```

with the following style sheet:

```scss
.sticky-heading-list {
  height: 100%;
  overflow: hidden;
  position: relative;

  // The main list
  & article.sticky-list-wrapper {
    height: 100%;
    overflow-x: hidden;
    overflow-y: scroll;

    // Section headers when "sticky"
    & section.sticky-section.sticky header {
      position: absolute;
      top: 0;
    }
  }
}
```

The problem is that when `<Waypoint>` calculates `_distanceToTopOfScrollableAncestor`, it uses `offsetParent`, which skips over the scrolling element `<article.sticky-list-wrapper>` in favor of `<div.sticky-heading-list>` (because it is positioned) which doesn't scroll and is 100% of the view port. So the return value is always `null`. 

By replacing the algorithm, it works independent of offsetParent, does not need to recurse up the DOM tree and eliminates a private method.

I'm concerned that the test on line 529, "does not call handlers until node becomes visible," does not actually test the right thing any more. I don't understand how deleting the parents position, then making it `relative` again makes it "visible". I admit that it's my own lack of understanding.

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -25,9 +25,9 @@ const scrollNodeTo = function(node, scrollTop) {
 describe('<Waypoint>', function() {
   beforeEach(() => {
     this.props = {
-      onEnter: jasmine.createSpy(),
-      onLeave: jasmine.createSpy(),
-      onPositionChange: jasmine.createSpy(),
+      onEnter: jasmine.createSpy('onEnter'),
+      onLeave: jasmine.createSpy('onLeave'),
+      onPositionChange: jasmine.createSpy('onPositionChange'),
       threshold: 0,
     };
 
@@ -519,6 +519,11 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
     beforeEach(() => {
       delete this.parentStyle.position;
+      this.topSpacerHeight = 200;
+
+      // The bottom spacer needs to be tall enough to force the Waypoint to exit
+      // the viewport when scrolled all the way down.
+      this.bottomSpacerHeight = 3000;
     });
 
     it('does not call handlers until node becomes visible', () => {
@@ -530,23 +535,28 @@ describe('<Waypoint>', function() {
 
       // now show it and we should get an onEnter
       node.style.position = 'relative';
-      scrollNodeTo(this.component, 10);
+      scrollNodeTo(this.component, 100);
       expect(this.props.onEnter).
         toHaveBeenCalledWith({
           currentPosition: Waypoint.inside,
-          previousPosition: Waypoint.invisible,
+          previousPosition: Waypoint.below,
           event: jasmine.any(Event),
         });
+    });
+  });
 
-      // now hide and we should see an onLeave
+  describe('when the scrollable parent is not displayed', () => {
+    it('calls the onLeave handler', () => {
+      this.component = this.subject();
+      const node = ReactDOM.findDOMNode(this.component);
       node.style.display = 'none';
-      scrollNodeTo(this.component, 20);
+      scrollNodeTo(this.component, 0);
       expect(this.props.onLeave).
-        toHaveBeenCalledWith({
-          currentPosition: Waypoint.invisible,
-          previousPosition: Waypoint.inside,
-          event: jasmine.any(Event),
-        });
+      toHaveBeenCalledWith({
+        currentPosition: Waypoint.invisible,
+        previousPosition: Waypoint.inside,
+        event: jasmine.any(Event),
+      });
     });
   });
 
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -516,35 +516,6 @@ describe('<Waypoint>', function() {
     });
   });
 
-  describe('when the scrollable parent does not have positioning', () => {
-    beforeEach(() => {
-      delete this.parentStyle.position;
-      this.topSpacerHeight = 200;
-
-      // The bottom spacer needs to be tall enough to force the Waypoint to exit
-      // the viewport when scrolled all the way down.
-      this.bottomSpacerHeight = 3000;
-    });
-
-    it('does not call handlers until node becomes visible', () => {
-      // initial render hidden
-      this.component = this.subject();
-      expect(this.props.onEnter).not.toHaveBeenCalled();
-      expect(this.props.onLeave).not.toHaveBeenCalled();
-      const node = ReactDOM.findDOMNode(this.component);
-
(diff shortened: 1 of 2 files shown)


spec/waypoint_spec.js
```diff
@@ -25,9 +25,9 @@ const scrollNodeTo = function(node, scrollTop) {
 describe('<Waypoint>', function() {
   beforeEach(() => {
     this.props = {
-      onEnter: jasmine.createSpy(),
-      onLeave: jasmine.createSpy(),
-      onPositionChange: jasmine.createSpy(),
+      onEnter: jasmine.createSpy('onEnter'),
+      onLeave: jasmine.createSpy('onLeave'),
+      onPositionChange: jasmine.createSpy('onPositionChange'),
```
trotzig: Neat! This will result in better error messages, right?

spec/waypoint_spec.js
```diff
@@ -530,23 +535,28 @@ describe('<Waypoint>', function() {
 
       // now show it and we should get an onEnter
       node.style.position = 'relative';
-      scrollNodeTo(this.component, 10);
+      scrollNodeTo(this.component, 100);
       expect(this.props.onEnter).
         toHaveBeenCalledWith({
           currentPosition: Waypoint.inside,
-          previousPosition: Waypoint.invisible,
+          previousPosition: Waypoint.below,
```
trotzig: The `invisible`ness was sort of a workaround due to limitations in the way we were calculating distance to the top. I think this might actually be better. Would you suggest removing that "position" after your change?

spec/waypoint_spec.js
```diff
@@ -530,23 +535,28 @@ describe('<Waypoint>', function() {
...
-        toHaveBeenCalledWith({
-          currentPosition: Waypoint.invisible,
-          previousPosition: Waypoint.inside,
-          event: jasmine.any(Event),
-        });
+      toHaveBeenCalledWith({
+        currentPosition: Waypoint.invisible,
+        previousPosition: Waypoint.inside,
+        event: jasmine.any(Event),
+      });
```
trotzig: Oh wait, I might have missed something. Looks like it will still apply.

trotzig:
This looks reasonable to me, and a good simplification. My only concern is performance. We know that `getBoundingClientRect` isn't super fast. We were already using it in one case (when `window` was the scrollable parent), and this commit adds another one. 

It would be interesting to hear if you ran any perf tests and what you found. It might be that the way we used to do things was in fact slower. 

Before merging, I'm going to ask @janpaul123 if he can review. He created #48, and we ended up having a discussion there about perf that might be of interest here.

trotzig:
Btw: thanks for taking the time to fix this!

src/waypoint.jsx
```diff
@@ -168,65 +168,44 @@ export default class Waypoint extends React.Component {
...
    *   visible portion of the scrollable parent. One of `POSITIONS.above`,
    *   `POSITIONS.below`, or `POSITIONS.inside`.
    */
   _currentPosition() {
-    const waypointTop =
-      this._distanceToTopOfScrollableAncestor(ReactDOM.findDOMNode(this));
-    if (waypointTop === null) {
-      // not visible
-      return POSITIONS.invisible;
-    }
```
janpaul123: This will behave slightly different now. For example, if you run `document.createElement('div').getBoundingClientRect();` you'll notice that all values are 0, which means that we'll return that it's inside the window (if that's the scrollable ancestor). If we want to keep the previous behaviour, we should add an explicit check for visibility.

spec/waypoint_spec.js
janpaul123: Yep!

spec/waypoint_spec.js
```diff
@@ -519,6 +519,11 @@ describe('<Waypoint>', function() {
   describe('when the scrollable parent does not have positioning', () => {
```
janpaul123: This section could be removed entirely, I think, as we won't rely on being positioned any more.

src/waypoint.jsx
janpaul123: Ah, never mind, I didn't read the line with `contextHeight === 0`. Nice!

src/waypoint.jsx
```diff
@@ -201,7 +201,7 @@ export default class Waypoint extends React.Component {
       return Waypoint.below;
     }
 
-    if (contextScrollTop > waypointTop + thresholdPx) {
+    if (waypointTop + thresholdPx < contextScrollTop) {
```
janpaul123: Could you also do that with line 200 (`if (waypointTop - thresholdPx > contextBottom)`)? See http://llewellynfalco.blogspot.com/2016/02/dont-use-greater-than-sign-in.html 😄

janpaul123:
This is great. Definitely simplifies the code, and removes the requirement to set `position: relative` on the scroll container.

Re performance @trotzig: if `getBoundingClientRect` was fast enough to use when the scrollable parent was `window`, then it's fast enough now (2 calls instead of 1 won't make a significant difference).

Thanks @kmayer! 👍

janpaul123:
Btw, my main comment before merging is https://github.com/brigade/react-waypoint/pull/82#discussion_r61665517, let's just remove that entire `describe`.

kmayer:
Before you merge, I need to do some more testing on Windows/IE ... I just did a quick QA and our loadMore callback fired once, then never again.

kmayer:
Okay, go ahead. IE needs 1? more pixel underneath the waypoint than Chrome does for it to appear "inside". Appending a spinner div resolves the issue and would be standard practice anyway.

trotzig:
Thanks all! I'll carve and publish a new version momentarily.

PR #92, opened 2016-06-04
waypoint HOC

bySabi:
This PR create waypoint HOC and implement Waypoint component like a wrapped of it.
Additionally:
* remove deprecated code
* support windows development
* enable use react-waypoint git repository like url package dependencie


build/npm/waypoint-hoc.js
```diff
@@ -218,6 +218,11 @@ var waypoint = function waypoint(Component) {
...
         }
       }
+
+      /**
+       * @return {Object} `height` and `top` of scrollable ancestor Rect.
+       */
+
     }, {
       key: '_scrollableAncestorHeightTop',
       value: function _scrollableAncestorHeightTop() {
```
trotzig: Files in the `build` folder are in the `.gitignore` file, so they shouldn't be commited.

spec/waypoint_spec.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import Waypoint from '../src/waypoint.jsx';
+import Waypoint from '../';
```
trotzig: I don't understand this change. What's being imported now?

package.json
```diff
@@ -1,53 +1,54 @@
...
+    "webpack": "^1.5.3"
+  },
+  "keywords": [
+    "react",
+    "component",
+    "react-component",
+    "scroll",
+    "onscroll"
+  ]
+}
```
trotzig: What changed here? The diff shows that the entire file changed. Are you using tabs or spaces?

src/waypoint-hoc.jsx
```diff
@@ -36,6 +40,9 @@ const waypoint = Component => {
     constructor(props) {
       super(props);
       this.state = { scrolled: {} };
+      this._handleScroll = (this.props.throttle ?
+        throttle(this._handleScroll, this.props.throttle) :
+        this._handleScroll).bind(this);
```
trotzig: What merited the throttle? I understand that it can improve performance, but I'd like to avoid introducing this complexity unless we are certain that we need it. The commit message needs to explain this. And I want this to be in a separate PR, complete with tests, a good commit message and possibly a benchmark script.

trotzig:
I appreciate you helping out @bySabi! This PR can't be merged as-is however. There are a couple of things you need to do:
- Help us out by explaining your changes better (a commit subject message isn't enough).
- Leave files in the `build` folder out of the mix (those are in the `.gitignore` file anyway)
- Split this PR into multiple ones: 
  - One for wrapping children (which I suggest you do with the children approach, not the HOC). 
  - One for adding the threshold prop, complete with a proper explanation why you did it and how it affects performance (with numbers)
  - One for adding more "DOM data" to callbacks, with an explanation why it's needed.

build/npm/waypoint-hoc.js
bySabi: We need build folder on repo for `npm install https://github.com/brigade/react-waypoint.git#master` cases.

spec/waypoint_spec.js
bySabi: it take the import from npm entrypoint (main),  `index.js` in this case.

package.json
bySabi: Sorry for that!! :-( .. I don´t why.
Maybe cause I clonned on a windows machine.
I use spaces.

src/waypoint-hoc.jsx
bySabi: Throttled scroll listeners is recommended all over the place, 
Ex:
* http://ejohn.org/blog/learning-from-twitter/
* http://benalman.com/code/projects/jquery-throttle-debounce/examples/throttle/

By default handleScroll is not throttled.
I agree this need more test and doc.

bySabi:
Not problem. 
I will split PR.

Issue #77, opened 2016-04-20
How about to add TopOffset and BottomOffset options for this tool?


bibobibo:
We have some sticky element in the scroll container, the sticky element height is const. In our case, the topLine is not really the scroll container top. so we need to add the topOffset and bottomOffset for the scroll container. Do you think it is necessary to add these two options for react-waypoint or make the threshold more flexible?

trotzig:
If I understand your issue correctly, you have something like this:

```
 -- scrollable container-----
|-- sticky bar --------------|
|--- waypoint ---------------|
|                            |
|                            |
|                            |
|                            |
|                            |
 ----------------------------
```

And you want to trigger the waypoint when it hits the sticky bar. Is that correct? 

I've actually had the same issue myself. I think I solved it at that time by hacking together something where the waypoint was inside a container with negative top position. That wasn't a great solution. I wonder if we could fix this by allowing the `threshold` prop to be a string as well? Then we could do something like 

```jsx
const STICKY_BAR_HEIGHT = '50px';
<Waypoint onEnter={handleEnter) threshold={STICKY_BAR_HEIGHT} />
```

I think that should be fairly easy to implement. Are you willing to take this on and write a PR?

lencioni:
I think the threshold should behave more like CSS values, and if it is a number, it is treated as px. This would allow us to use `threshold={100}` to mean 100px, and `threshold='100px'`/`threshold='100%'` are a lot more intuitive.

trotzig:
That sounds like a good idea @lencioni. 

Would `20%` be the same as `0.2` (as we have today)?

If we decide to implement support for these units, I suggest we continue to support the current setup, but log a deprecation warning if you do. So `threshold={0.2}` would be translated to `20%` and generate a warning of the following kind:

```
Support for floating point threshold is going away soon. Please use `'20%'` instead of `0.2`. 
```

It gets a little tricky if you use `threshold={1}` however. Hmm, maybe we should just rename the prop and throw an error instead? I guess proper release notes could work too, but in my experience people don't read them too carefully, and a change to the behavior of a prop can have bad consequences.

lencioni:
Yes `0.2` would become `'20%'`. I think pairing this with a major version change and a codemod that will fix the problems for you should be good.

jmeas:
[...]

PR #94, opened 2016-06-06
Replace threshold with topOffset and bottomOffset

jmeas:
This is a WIP to resolve https://github.com/brigade/react-waypoint/issues/77

It has two notable changes:

1. `threshold` has been split into `offsetTop` and `offsetBottom`
2. The new options accept strings and numbers. Strings can be of the form `"100"`, `"100px"` or `"20%"`. Actual numbers, like `20`, are assumed to be pixel values

---

I haven't added tests 'cause I figure there might be feedback on the implementation. Once the implementation is :+1:'d I'll update the tests.

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -10,19 +10,33 @@ const POSITIONS = {
 
 const propTypes = {
   debug: PropTypes.bool,
-  // threshold is percentage of the height of the visible part of the
-  // scrollable ancestor (e.g. 0.1)
-  threshold: PropTypes.number,
   onEnter: PropTypes.func,
   onLeave: PropTypes.func,
   onPositionChange: PropTypes.func,
   fireOnRapidScroll: PropTypes.bool,
   scrollableAncestor: PropTypes.any,
-  throttleHandler: PropTypes.func
+  throttleHandler: PropTypes.func,
+  // `topOffset` can either be a number, in which case its a distance from the
+  // top of the container in pixels, or a string value. Valid string values are
+  // of the form "20px", which is parsed as pixels, or "20%", which is parsed
+  // as a percentage of the height of the containing element.
+  // For instance, if you pass "-20%", and the containing element is 100px tall,
+  // then the waypoint will be triggered when it has been scrolled 20px beyond
+  // the top of the containing element.
+  topOffset: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  // `bottomOffset` is like `topOffset`, but for the bottom of the container.
+  bottomOffset: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
 };
 
 const defaultProps = {
-  threshold: 0,
+  topOffset: '0px',
+  bottomOffset: '0px',
   onEnter() {},
   onLeave() {},
   onPositionChange() {},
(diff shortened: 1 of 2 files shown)


src/waypoint.jsx
```diff
@@ -32,6 +45,20 @@ function debugLog() {
   console.log(arguments); // eslint-disable-line no-console
 }
 
+function parseOffsetAsPixels(str) {
```
jmeas: lmk if you'd rather these be instance props on the components @trotzig

src/waypoint.jsx
```diff
@@ -202,23 +253,24 @@ export default class Waypoint extends React.Component {
...
+    const bottomOffsetPx = this._computeBottomOffsetPx(contextHeight);
     const contextBottom = contextScrollTop + contextHeight;
 
     if (contextHeight === 0) {
       return Waypoint.invisible;
     }
 
-    if (contextScrollTop <= waypointTop + thresholdPx &&
-        waypointTop - thresholdPx <= contextBottom) {
+    if (contextScrollTop <= waypointTop - topOffsetPx &&
```
jmeas: I'll likely need to adjust the relational operators here. Getting unexpected results when just flipping the `+` to `-`.

src/waypoint.jsx
```diff
@@ -18,10 +18,23 @@ const propTypes = {
   onPositionChange: PropTypes.func,
   fireOnRapidScroll: PropTypes.bool,
   scrollableAncestor: PropTypes.any,
+  // `topOffset` can either be a number, in which case its a distance from the
+  // top of the container, or a percentage, in which case it's a percentage of
+  // the height of the visible part of the scrollable ancestor (e.g. 0.1)
```
trotzig: The `e.g. 0.1` part is confusing.

src/waypoint.jsx
trotzig: No, this is better. :)

trotzig: On second thought, maybe they should be. I don't think computing these over and over affect performance too much, but it's an easy change.

trotzig:
This looks good! You may go ahead and fix tests. 👍

src/waypoint.jsx
jmeas: I'll move em in and we can see how it looks 🙂

src/waypoint.jsx
jmeas: I'll update the comment.

jmeas:
> This looks good! You may go ahead and fix tests

It might take a few days, but it's on my todo list!

src/waypoint.jsx
jmeas: > I don't think computing these over and over affect performance too much, but it's an easy change.

As an aside, I don't think either approach causes anything to be computed or created more than the other.

Most JS module loaders cache the modules, so these will be created the first time that the lib is required and never again. And if we move it into the class, they'll be defined on the prototype, also just once. So it should just be a matter of preference, which could be an aesthetic preference or one based on whether it makes sense to expose these to consumers of the lib.

trotzig: Oh, I thought you were talking about memoizing the value. I was thinking we could store the computed value in an instance variable and only call the method(s) once. But I realize now that we can't do that because of the `contextHeight` being variable. Sorry for the confusion.

jmeas: Ah, okay! I misunderstood - sorry 🙂

jmeas:
It's been some time, but I've rebased this off master and made the existing tests pass (with a small tweak to the current test for threshold). I still need to write the rest of the offset tests (offset top + pixel values for offset), but I figured I'd post this update :v:

trotzig:
Cool, this is coming together nicely.

jmeas:
@trotzig I updated the JSDoc comments and added tests for all of the different ways to specify the bottom offset. I'd love to get any feedback you have on the JSDoc comments.

Would you also like to see analogous tests for the top offset prop?

trotzig:
Sorry for the delay. Let me have a look at this tomorrow when my eyes are less tired. 🌃

jmeas:
Sure thing. Thanks!

src/waypoint.jsx
```diff
@@ -184,6 +198,62 @@ export default class Waypoint extends React.Component {
   }
 
   /**
+   * @param {string|number} offset Blah blah blah
+   * @param {number} contextHeight Blah blah blah
```
trotzig: lol, you might want to change this description a little. I think just removing the blah-blah is fine too.

src/waypoint.jsx
```diff
@@ -206,23 +276,26 @@ export default class Waypoint extends React.Component {
       debugLog('scrollableAncestor height', contextHeight);
       debugLog('scrollableAncestor scrollTop', contextScrollTop);
     }
-    const thresholdPx = contextHeight * this.props.threshold;
+
+    const {bottomOffset, topOffset} = this.props;
```
trotzig: No big deal, but we've mostly used spaces inside curlies: 

```js
const { bottomOffset, topOffset } = ...
```

src/waypoint.jsx
jmeas: Hah. Oops. I missed this blah blah. I'll update.

src/waypoint.jsx
jmeas: Consistency is good – I'll fix this up real quick.

trotzig:
All good. If it's not a big deal, we should probably have a simple test for topOffset as well. Don't flesh it out like you did for bottomOffset though. We're probably fine with a single test that checks for onEnter being called.

jmeas:
I'll put that together tomorrow.

jmeas:
Woops – took a few days longer than expected, but this has been updated. The "blah blah"s are gone, and there's a new test for `topOffset`. I also squashed the commits.

@trotzing, the current setup of the block of the tests that the `bottomOffset` tests made it easier to test for `onLeave` than `onEnter`, so I went ahead and did that.

Lmk if there's anything else I should do/change here :)

trotzig:
Thanks! I'm releasing this as 3.0.0 (going from 2.0.3, since this is a breaking change). Stay tuned.

jmeas:
Rad. Thanks 🙂

PR #103, opened 2016-06-20
Use in-publish to ensure prepublish is only run at right time

jmeas:
The `prepublish` script has a bad name, I think, 'cause most people use it for things they only want to run right before publishing to npm, yet the script also runs on `npm install`. It looks like this lib only intends for it to be run before publish (correct me if I'm wrong!), so this PR adds the [`in-publish`](https://github.com/iarna/in-publish) lib to filter how frequently the task is run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint": "eslint . --ext .js,.jsx",
     "testonly": "karma start",
     "test": "npm run lint && npm run testonly",
-    "prepublish": "npm run build-npm"
+    "prepublish": "in-publish && npm run build-npm || not-in-publish"
   },
   "author": "Brigade Engineering",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -33,17 +33,18 @@
...
     "karma-chrome-launcher": "^0.2.0",
     "karma-cli": "^0.1.0",
     "karma-firefox-launcher": "^0.1.6",
     "karma-jasmine": "^0.3.6",
     "karma-webpack": "^1.7.0",
+    "mkdirp": "^0.5.1",
     "react": "^15.0.0",
     "react-dom": "^15.0.0",
     "rimraf": "^2.5.2",
-    "mkdirp": "^0.5.1",
```
jmeas: This was moved because this file was manually edited at some point. Using the CLI for installing deps keeps them in alphabetical order.

trotzig: Thanks!

trotzig:
Any ideas why Travis isn't happy with this?

jmeas:
Hm, not sure. I'll take a look today.

jmeas:
Fixed :tada:

I had forgotten to include the `not-in-publish` command.

trotzig:
Thanks!

jmeas:
Thank you! :v:

PR #104, opened 2016-06-20
Add scrollspy as example for how this lib can be used

jmeas:
I'm using this library to implement [scrollspying](http://v4-alpha.getbootstrap.com/components/scrollspy/), so I thought it might be good to call it out in the README. Perhaps surprisingly, this library is by far the best lib I've found to implement Scrollspies (even better than React scrollspy-specific libraries, imo), which is why I think it's a good idea to mention it in the README.

NBD if you don't think it's a good example, though :)


trotzig:
This is great, we use it ourselves for scrollspying.

Issue #107, opened 2016-07-06
Container positions


thomassuckow:
It would be useful to have the following exposed in the events:

waypointTop
contextScrollTop + topOffsetPx
contextBottom - bottomOffsetPx

The latter two might be: viewportTop, viewportBottom respectively.

lencioni:
Can you help us understand your use-case?

thomassuckow:
I am using react-waypoint to render very large scrollable regions. Think infinite scroll except (at least for now) it isn't infinite. Because there is a lot of complexity to what it being displayed I have broken it into "pages" that get rendered only when within the viewport. When one scrolls up or down pages get inserted or removed.

The unfortunate thing is the pages vary in size and can change. The waypoints make it easy to track the movement of the pages with respect to the viewport and I can add or remove pages as one enters / leaves. But if you scroll quickly I can at best add a page (and a waypoint), get another event that it wasn't enough to fill the viewport and add another. This continues until the page catches up with the scroll position which takes seconds if not minutes. If I know the pixel position of the waypoint relative to the viewport I can skip pages and go right to the page that should be in view.

I am working on a PR, after I posted this issue I decided there wasn't really any effective way to do what I wanted without modding react-waypoint or using a different library. It is a pretty simple change but requires modifying half the tests.

jmeas:
@thomassuckow is that open source? I get the gist of it, but I'm curious to see the code.

bySabi:
@thomassuckow do you see the HoC proposal PR thread: https://github.com/brigade/react-waypoint/pull/105 ??
Maybe you can create your own solution with the HoC.

There you can ask anything.

thomassuckow:
@jmeas Unfortunately the infinite scroll implementation has not been open sourced. It isn't finished either... I finished the changes to this library last night but haven't had a chance yet to integrate and make sure it actually solves my problems.

@bySabi I did see the HoC thread. It appears to have basically the same functionality as the current react-waypoint component though has the advantage of being able to wrap a component without adding more DOM. It would probably be quite similar to add the functionality to the HoC version. The HoC version would probably also want a waypointBottom value since it wraps a component occupying space.

[...]

PR #109, opened 2016-07-13
Feature/bounds

thomassuckow:
Resolve adding bounds values as described in #107 

Also resolves an issue with findDOMNode "Uncaught Invariant Violation: getNodeFromInstance"

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -148,7 +148,8 @@ export default class Waypoint extends React.Component {
    *   called by a React lifecyle method
    */
   _handleScroll(event) {
-    const currentPosition = this._currentPosition();
+    const bounds = this._getBounds();
+    const currentPosition = this._currentPosition(bounds);
     const previousPosition = this._previousPosition || null;
     if (this.props.debug) {
       debugLog('currentPosition', currentPosition);
@@ -167,6 +168,9 @@ export default class Waypoint extends React.Component {
       currentPosition,
       previousPosition,
       event,
+      waypointTop: bounds.waypointTop,
+      viewportTop: bounds.viewportTop,
+      viewportBottom: bounds.viewportBottom,
     };
     this.props.onPositionChange.call(this, callbackArg);
 
@@ -188,11 +192,17 @@ export default class Waypoint extends React.Component {
         currentPosition: POSITIONS.inside,
         previousPosition,
         event,
+        waypointTop: bounds.waypointTop,
+        viewportTop: bounds.viewportTop,
+        viewportBottom: bounds.viewportBottom,
       });
       this.props.onLeave.call(this, {
         currentPosition,
         previousPosition: POSITIONS.inside,
         event,
+        waypointTop: bounds.waypointTop,
+        viewportTop: bounds.viewportTop,
+        viewportBottom: bounds.viewportBottom,
       });
     }
   }
@@ -253,12 +263,7 @@ export default class Waypoint extends React.Component {
     }
   }
 
-  /**
-   * @return {string} The current position of the waypoint in relation to the
-   *   visible portion of the scrollable parent. One of `POSITIONS.above`,
-   *   `POSITIONS.below`, or `POSITIONS.inside`.
-   */
-  _currentPosition() {
+  _getBounds() {
     const waypointTop = ReactDOM.findDOMNode(this).getBoundingClientRect().top;
     let contextHeight;
     let contextScrollTop;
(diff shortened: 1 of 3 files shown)


src/waypoint.jsx
```diff
@@ -264,16 +269,15 @@ export default class Waypoint extends React.Component {
...
     let contextHeight;
     let contextScrollTop;
     if (this.scrollableAncestor === window) {
       contextHeight = window.innerHeight;
       contextScrollTop = 0;
     } else {
       contextHeight = this.scrollableAncestor.offsetHeight;
-      contextScrollTop = ReactDOM
-        .findDOMNode(this.scrollableAncestor)
+      contextScrollTop = this.scrollableAncestor
```
jmeas: Worth noting that this would be a breaking change

jmeas:
> Also resolves an issue with findDOMNode "Uncaught Invariant Violation: getNodeFromInstance"

What's this issue?

src/waypoint.jsx
thomassuckow: I don't understand how this changes anything.
```this.scrollableAncestor``` is a dom element so
```ReactDOM.findDOMNode(this.scrollableAncestor) === this.scrollableAncestor```

**Comment by thomassuckow** [comment]
https://github.com/apollostack/meteor-starter-kit/issues/16 describes the issue I was seeing while working on these changes. For some reason building react-waypoint on my machine caused ReactDOM.findDOMNode to throw an invariant exception.

**Comment by thomassuckow** [comment]
I would normally separate them, but the changes overlap so they would have to be submitted one at a time. I did pull the third commit (No forced layout on mount/update) though because it isn't strictly necessary and breaks the tests.

File: src/waypoint.jsx (hunk 1 above)
Comment (jmeas) [inline src/waypoint.jsx]: Ah, you're right. I was thinking that passing a component instance in as a `scrollableAncestor` would work (by leveraging `findDOMNode`), but I took another look at the source and saw that it would error earlier up in the code.

Ignore me! 🙈

File: src/waypoint.jsx [hunk 2]
```diff
@@ -188,11 +192,17 @@ export default class Waypoint extends React.Component {
...
+        viewportTop: bounds.viewportTop,
+        viewportBottom: bounds.viewportBottom,
       });
       this.props.onLeave.call(this, {
         currentPosition,
         previousPosition: POSITIONS.inside,
         event,
+        waypointTop: bounds.waypointTop,
+        viewportTop: bounds.viewportTop,
+        viewportBottom: bounds.viewportBottom,
```
trotzig: I think you could use an object spread here (and in two places above) to avoid having to repeat the bounds properties:

```
this.props.onLeave.call(this, {
  currentPosition,
  previousPosition: POSITIONS.inside,
  event,
  ...bounds,
});
```

src/waypoint.jsx
trotzig: Good catch!

src/waypoint.jsx
```diff
@@ -328,7 +332,7 @@ export default class Waypoint extends React.Component {
   render() {
     // We need an element that we can locate in the DOM to determine where it is
     // rendered relative to the top of its context.
-    return <span style={{fontSize: 0}} />;
+    return <span ref={this.refElement} style={{fontSize: 0}} />;
```
trotzig: Thanks for not using an anonymous function to handle the ref setting.

trotzig:
Thanks @thomassuckow, this all looks good to me. Good call on pulling the third commit (makes this PR easier to merge), although I'd love to see a separate PR for that!

I'll let you decide if you want to take action on my comment to DRY up src/waypoint.jsx a little before merging.

src/waypoint.jsx
thomassuckow: I could, but I would need to add a few babel plugins. If you would prefer that syntax I can do so.
https://babeljs.io/docs/plugins/transform-object-rest-spread/

thomassuckow:
DRY:
- [ ] Object rest spread

Unless I am mistaken the above is the only DRYing change, if I am wrong point it out. It is also dependant on whether this project wants to add more babel plugins or if there is some reason not to.

trotzig:
Ah, I thought we had that wired up already. Let's merge this without it. We can change that later on.

src/waypoint.jsx
lencioni: This is fine, but I'm not sure it really makes much difference either way. My preference would be to use the arrow function here to reduce unnecessary indirection.

lencioni:
rest/spread is only a stage 2 proposal right now. I think we should wait until it hits stage 3 to bring it in.

https://github.com/sebmarkbage/ecmascript-rest-spread

src/waypoint.jsx
thomassuckow: I've never been able to prove from looking at the react source that an anonymous function on `ref` does or does not cause a the virtual DOM to be synced to the real DOM. You cannot inspect `ref` in shouldComponentUpdate which would imply to me that is does not cause it to be synced but I was never able to definitively determine that.

eslint if configured to prohibit anonymous functions in `render` will still complain if it is `ref`, but maybe it is just too strict.

trotzig:
I included this change in v3.1.0 (just released). Thanks for the contribution @thomassuckow!

PR #114, opened 2016-08-03
Passing props from constructor() to super() method

kamotos:
Not passing props through super() breaks Waypoint usage in [preact](https://github.com/developit/preact).

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -53,8 +53,8 @@ function debugLog() {
  * Calls a function when you scroll to the element.
  */
 export default class Waypoint extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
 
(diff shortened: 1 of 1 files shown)


lencioni:
Thanks @kamotos for your contribution! Can you be more specific about how it breaks Waypoint usage in preact?

kamotos:
As `props` is not passed to the parent _constructor_, Waypoint is instantiated without the user defined props. which results in `this.props == undefined` and that throws an error at the first props access, eventually "Cannot read property `scrollableParent` of undefined."

Let me know if I need to provide further explanation.

trotzig:
Thanks for the contribution @kamotos! This looks fair to me.

trotzig:
I just published 3.1.1 with this change.

jmeas:
Thanks @kamotos 🎉 🎂 💃

PR #123, opened 2016-10-18
Manage a single window resize event instead of one per component

lencioni:
I am looking into improving the performance of react-waypoint. We can be
a little more mindful of how many event listeners we end up creating by
only managing a single event listener. Using Chrome's timeline tool, I
have verified that this type of optimization is significant when there
are a lot of event listeners on the page.

Since we always set up a window resize event, this seemed like a good
place to start for this optimization. I chose a Set, since that seemed
like the easy and nice way to go, but it will require that older
browsers have a polyfill.

If we like this approach, I plan to do something similar for the scroll
event listeners. These are a little more complicated since they could be
attached to any scrollable container, instead of just the window. One
option would be to use a Map to manage a Set of event listeners for each
scrollable container. It might also be worth exploring using a single
top-level event listener instead of one per scrollable container, which
should be accomplishable via the Map/Set approach.

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -23,6 +23,34 @@ function debugLog() {
   console.log(arguments); // eslint-disable-line no-console
 }
 
+const resizeEventHandlers = {
+  size: 0,
+  index: 0,
+};
+
+function handleResize(event) {
+  resizeEventHandlers.forEach(handler => handler(event));
+}
+
+function addResizeEventListener(listener) {
+  if (resizeEventHandlers.size === 0) {
+    window.addEventListener('resize', handleResize);
+  }
+
+  resizeEventHandlers[resizeEventHandlers.index++] = listener;
+  resizeEventHandlers.size++;
+  return resizeEventHandlers.index;
+}
+
+function removeResizeEventListener(index) {
+  delete resizeEventHandlers[index];
+  resizeEventHandlers.size--;
+
+  if (resizeEventHandlers.size === 0) {
+    window.removeEventListener('resize', handleResize);
+  }
+}
+
 /**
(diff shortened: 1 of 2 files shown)


src/waypoint.jsx
```diff
@@ -23,6 +23,30 @@ function debugLog() {
   console.log(arguments); // eslint-disable-line no-console
 }
 
+let isResizeEventListenerAdded = false;
```
lencioni: I realized that I don't actually need this, because I can use `size` from the Set. I'll fix this if folks are generally okay with this direction. I forgot to set it back to false in the remove event listener function anyway.

trotzig:
This is a good start! Was your plan for polyfilling to bundle it with react-waypoint? It wouldn't be too hard to use `splice` and `indexOf` too, which would have more broad support. Also, how do you feel about making `addResizeEventListener` return a "subscription" of sorts, that you can use to remove it? 

```jsx
componentDidMount() {
  this._resizeSubscription = addResizeEventListener(this._someFunc);
}

componentWillUnmount() {
  this._resizeSubscription.remove();
}

```

lencioni:
My plan was to not polyfill it, but to list it in the readme as a dependency. I want to avoid the `indexOf` route because that requires a full scan of the array every time. Maybe I'll use an object with your subscription suggestion instead. That should be both fast and well supported.

lencioni:
@trotzig I've updated this PR with an object-based approach. PTAL

trotzig:
Cool, thanks for addressing! Merge at will.

lencioni:
Merged. This afternoon, I plan to expand this approach to the scroll event listeners.

jmeas:
👌nice

PR #125, opened 2016-10-19
Update devDependencies

lencioni:


.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: node_js
+sudo: false
 node_js:
-  - '0.10'
+  - '6'
+  - '5'
```
jmeas: My understanding of Node's LTS support is that v5 isn't being maintained – is it worth testing against, you think?

lencioni: I don't think it matters much either way. We can remove it if you like.

PR #128, opened 2016-10-20
Avoid leaking event listeners

lencioni:
In c6711294 I made a change to waypoint that makes it so it sets up
fewer event listeners. Unfortunately, I did this in a way that causes
the added event listener to never be removed, as @trotzig pointed out at

  https://github.com/brigade/react-waypoint/commit/c6711294#commitcomment-19498628

This is happening because every time we call .bind() on a function, we
get a new object, which the browser is unable to use to remove the
previous function. The solution is to bind the function once, and keep
track of it so we can reuse the reference when we want to remove the
event listener.

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -63,6 +63,7 @@ class TargetEventHandlers {
         size: 0,
         index: 0,
         handlers: {},
+        handleEvent: undefined,
       };
     }
     return this.events[eventName];
@@ -86,9 +87,11 @@ class TargetEventHandlers {
     const eventHandlers = this.getEventHandlers(eventName);
 
     if (eventHandlers.size === 0) {
+      eventHandlers.handleEvent = this.handleEvent.bind(this, eventName);
+
       this.target.addEventListener(
         eventName,
-        this.handleEvent.bind(this, eventName),
+        eventHandlers.handleEvent,
         EVENT_OPTIONS
       );
     }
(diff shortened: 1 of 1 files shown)


src/waypoint.jsx
```diff
@@ -107,9 +110,11 @@ class TargetEventHandlers {
     if (eventHandlers.size === 0) {
       this.target.removeEventListener(
         eventName,
-        this.handleEvent.bind(this, eventName),
+        eventHandlers.handleEvent,
         EVENT_OPTIONS
       );
+
+      eventHandlers.handleEvent = undefined;
```
trotzig: That was unclear... I mean how would you feel about `delete eventHandlers.handleEvent`?

lencioni: That seems reasonable. Feel free to change it. I'm going to leave it as is for now.

trotzig:
When I spotted this, I started thinking about ways to test it. First I tried making an `afterEach` throw if there were listeners remaining on `window`. But it turns out it's really hard to get a list of event listeners for a node. Chrome has `getEventListeners(node)` but it only works from the console. Any ideas?

lencioni:
You could add spies on add/remove event listener and write tests to verify that they were called correctly. The problem is that you could have the same bugs in the test that we had in the code.

Another alternative would be to make the handleEvent function that is added/removed a spy, and then do the behavior to cause it to be added and removed, and then do something that would trigger the event and see if it is called. Although, I'm not entirely sure that would have caught this bug.

PR #132, opened 2016-10-29
Add yarn installation method

Kerumen:


trotzig:
Hi @Kerumen, and thanks for contributing to react-waypoint! I hadn't heard of yarn before, thanks for introducing this to me/us. Can you give a little bit more context on why you think we should have this in the README. It could be argued that yarn users should be aware of the fact that they can install npm modules.

Kerumen:
yarn uses the same registry as npm so it's transparent for the users. You can [read more here](https://code.facebook.com/posts/1840075619545360).

PR #135, opened 2016-11-09
Extract event handling code to separate package

lencioni:
I want to use the event handling code that I wrote for react-waypoint in
some different projects, so I decided to extract it to its own package,
generalize it some, and write some tests. This commit completes the
extraction by removing the code from react-waypoint, adding the
dependency, and updating the usage to match the new API.

https://github.com/lencioni/consolidated-events

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import { addEventListener, removeEventListener } from 'consolidated-events';
 
 const POSITIONS = {
   above: 'above',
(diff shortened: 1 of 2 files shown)


src/waypoint.jsx
```diff
@@ -252,16 +138,18 @@ export default class Waypoint extends React.Component {
       debugLog('scrollableAncestor', this.scrollableAncestor);
     }
 
-    this.scrollEventListenerId = addEventListener(
+    this.scrollEventListenerHandle = addEventListener(
```
jmeas: a thing to consider is that folks reading this could conceivably mistake this for a `window.addEventListener` due to the shadowing. I don't think anything needs changing here, but it's something I thought of while reading through the changes.

PR #136, opened 2016-11-10
Delay initial calling of handleScroll when mounting

lencioni:
At Airbnb we are using react-waypoint to do some lazy image loading. We
noticed a problem on a page that has three images that are initially
rendered within the viewport and use waypoint to start the image loading
process. Sometimes one or two of the images don't start loading until
after you scroll a little, even if they are initially rendered within
the viewport.

Our hypothesis is that `this._ref` is not ready yet at the time this
function is called, which is guarded against in the handleScroll method.
We believe that by delaying the initial execution of this function, the
browser will have time to wire up the ref and our waypoints will
appropriately trigger when they are initially rendered.

We have tested this out in a development environment, and it seems to
resolve the issue for us.

cc: @nicksorrentino

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -24,6 +24,7 @@ const scrollNodeTo = function(node, scrollTop) {
 
 describe('<Waypoint>', function() {
   beforeEach(() => {
+    jasmine.clock().install();
     spyOn(console, 'log');
     this.props = {
       onEnter: jasmine.createSpy('onEnter'),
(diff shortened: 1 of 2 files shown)


trotzig:
Looks good.

nicksorrentino:
Thanks @lencioni

trotzig:
For context, here are a few things @lencioni and I talked to outside of this thread: 
- Because we're adding a defer/setTimeout, this could be a breaking change. People might have tests for instance expecting a waypoint to have fired as part of rendering. 
- We believe the breaking change to be small enough to go out as a patch version anyway. There's no change in the waypoint's interface.

jmeas:
Seems like sound reasoning to me :+1:

PR #137, opened 2016-11-12
Add horizontal scrolling

### LINKED ISSUE #66: Horizontal scrolling
Would it be possible to enable waypoints on horizontal scrolling nodes?

I have a vertical list of items that I want to load more of while scrolling, but on smaller screens it switches to a horizontal scrolling implementation.  However, the waypoint loads everything initially.

agilgur5:
Implements #66 

To explain some of my rationale, as I was writing this I quickly realized that little of the internal logic actually needs to change to support horizontal scrolling (approx. 7 lines).
Instead of creating new `POSITIONS` and forking a bunch of functions and creating more variables (e.g. `viewportLeft`, `waypointLeft`, etc), it would actually make more sense to abstract all the code into things like `viewportEdge1`, `contextSize`, `contextScrollEdge1`, `POSITIONS.outside1`, `POSITIONS.outside2`, but as I started renaming everything, I realized that would be a major breaking API change.
Ideally, I think there should be some internal implementation like that, and then a wrapper component can just convert props like `bottomOffset` or `leftOffset` to `edge1Offset`, but for now, this is a very quick fix that adds a big, requested feature (that my organization also needs) without making any API changes. It's also the most straightforward approach to adding this, it would seem, as logic like the above creates a really confusing naming scheme while coding due to inexplicit and abstract names.

With regard to testing, I wasn't sure how to add a test for this without rewriting a lot of the test code. As this only adds a few extra lines that should be tested, I don't believe all the tests need to be redone (the logic is the same, just in a different direction), but a test to check that it's measuring horizontal distance when `horizontal` is set would be good. All the testing code specifically measures vertical distance though, so I again ran into the problem of rewriting a lot of the test code for a very small change.

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -11,6 +11,7 @@ const POSITIONS = {
 const defaultProps = {
   topOffset: '0px',
   bottomOffset: '0px',
+  horizontal: false,
   onEnter() {},
   onLeave() {},
   onPositionChange() {},
@@ -207,10 +208,12 @@ export default class Waypoint extends React.Component {
       }
 
       const style = window.getComputedStyle(node);
-      const overflowY = style.getPropertyValue('overflow-y') ||
-        style.getPropertyValue('overflow');
+      const overflowDirec = this.props.horizontal ?
+        style.getPropertyValue('overflow-x') :
+        style.getPropertyValue('overflow-y');
+      const overflow = overflowDirec || style.getPropertyValue('overflow');
 
-      if (overflowY === 'auto' || overflowY === 'scroll') {
+      if (overflow === 'auto' || overflow === 'scroll') {
         return node;
       }
     }
@@ -292,16 +295,21 @@ export default class Waypoint extends React.Component {
   }
 
   _getBounds() {
-    const waypointTop = this._ref.getBoundingClientRect().top;
+    const horizontal = this.props.horizontal;
+    const waypointTop = horizontal ? this._ref.getBoundingClientRect().left :
+      this._ref.getBoundingClientRect().top;
 
     let contextHeight;
     let contextScrollTop;
     if (this.scrollableAncestor === window) {
-      contextHeight = window.innerHeight;
+      contextHeight = horizontal ? window.innerWidth : window.innerHeight;
       contextScrollTop = 0;
     } else {
-      contextHeight = this.scrollableAncestor.offsetHeight;
-      contextScrollTop = this.scrollableAncestor.getBoundingClientRect().top;
+      contextHeight = horizontal ? this.scrollableAncestor.offsetWidth :
+        this.scrollableAncestor.offsetHeight;
+      contextScrollTop = horizontal ?
+        this.scrollableAncestor.getBoundingClientRect().left :
+        this.scrollableAncestor.getBoundingClientRect().top;
     }
 
     if (this.props.debug) {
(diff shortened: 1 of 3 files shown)


trotzig:
Awesome @agilgur5! I'll be traveling for a few days, but I'll have a look at it as soon as I can.

src/waypoint.jsx
```diff
@@ -292,16 +295,21 @@ export default class Waypoint extends React.Component {
   }
 
   _getBounds() {
-    const waypointTop = this._ref.getBoundingClientRect().top;
+    const horizontal = this.props.horizontal;
+    const waypointTop = horizontal ? this._ref.getBoundingClientRect().left :
```
trotzig: Perhaps we should rename this variable to `waypointStart`?

src/waypoint.jsx
```diff
@@ -292,16 +295,21 @@ export default class Waypoint extends React.Component {
...
+      this._ref.getBoundingClientRect().top;
 
     let contextHeight;
     let contextScrollTop;
     if (this.scrollableAncestor === window) {
-      contextHeight = window.innerHeight;
+      contextHeight = horizontal ? window.innerWidth : window.innerHeight;
```
trotzig: Then this could be `contextSize`.

src/waypoint.jsx
```diff
@@ -292,16 +295,21 @@ export default class Waypoint extends React.Component {
...
       contextScrollTop = 0;
     } else {
-      contextHeight = this.scrollableAncestor.offsetHeight;
-      contextScrollTop = this.scrollableAncestor.getBoundingClientRect().top;
+      contextHeight = horizontal ? this.scrollableAncestor.offsetWidth :
+        this.scrollableAncestor.offsetHeight;
+      contextScrollTop = horizontal ?
```
trotzig: And `contextScrollStart` here.

trotzig:
I think it was the right decision to keep this change small at this time. We can always refactor/rename in the future. Since vertical scrolling makes up the vast majority of use-cases, naming things based on it is probably fine. Names will be awkward for horizontal waypoints, but most people will only use `onEnter` anyway, so they're not exposed to the different positions a waypoint can be in. 

Before we merge this, it would be good to add at least a smoke-test for horizontal scroll. I agree that we don't need to fork tests too much. Preferably, I'd like to see the following tests: 

- a horizontal waypoint with `window` as its scrollable ancestor.
- a horizontal waypoint with a DOM element as its scrollable ancestor. 

Simple tests that verify that `onEnter` is triggered on mount are enough. I don't think we need to test the actual scrolling. 

Thanks for the contribution!

src/waypoint.jsx
```diff
@@ -11,6 +11,7 @@ const POSITIONS = {
 const defaultProps = {
   topOffset: '0px',
   bottomOffset: '0px',
```
lencioni: It seems like we should rename these props at some point to be `offsetBefore` and `offsetAfter`. If we do that, it would be good to keep these around with a deprecation warning for one major release.

lencioni:
@nicksorrentino what do you think about this change?

nicksorrentino:
@lencioni this is looking good, we would most likely using this on our horizontal scroll work we have.  I will pull the branch and test it out

agilgur5:
@trotzig I made the requested changes to the README and added one test, but wasn't entirely sure how to make the scrollable ancestor changes :/

trotzig:
Thanks @agilgur5! I made some amendments in 74dbd00f60c968d9 before merging. I'll carve a new release asap.

trotzig:
This was released in `4.1.0`.

PR #139, opened 2016-11-17
Improve publishing process

lencioni:
I recently added a few npm scripts to help making the release process go
smoother. Unfortunately, they caused a little bit of confusion because
some folks were still trying to use the built-in `npm version` command
instead of the new `npm run version:foo` scripts.

I poked around npm documentation a bit and discovered that we can set
things up so that `npm version` does what we want. The magic is done by
moving the command flag to avoid the automatic creation of the git tag
out of the script and into a .npmrc file.

The result of this is now `npm version major|minor|patch` will manage
the full release process for you.

To: @trotzig @ljharb

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+git-tag-version=false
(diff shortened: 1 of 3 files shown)


ljharb:
Very interesting. It'd be really cool to combine all this stuff into one package that can be run preversion and/or postversion, so configuring this stuff can be easier.

trotzig:
Thanks Joe! I'm sorry for being on the defensive side in our off-github discussion. I'm glad it resulted in this. ❤️

Issue #129, opened 2016-10-20
Uses for `throttleHandler` prop


lencioni:
Related: #73, #98

In v4, we made some performance improvements to add fewer event listeners, and to also use passive event listeners. With these changes I'm wondering if the `throttleHandler` prop is still useful. In our codebase, I've witnessed folks using it even when it wasn't useful, so I am wondering if it adds more confusion and uncertainty about the performance characteristics of this component than it is worth. If it isn't useful, or if the same thing can be accomplished via composition, I would like to remove this prop from this component to simplify its API.

Can folks here who are using `throttleHandler` check out the performance characteristics of v4.0.2+ without the `throttleHandler` prop and please post a comment on this issue with what you've found? I'd really like to have a better understanding of how you are using this prop and what benefits you are seeing from its use.

trotzig:
I'm in favor of removing it unless we can provide some numbers telling us it's actually useful. We're not using it ourselves despite some pages having 20-30 waypoints. 

@bySabi could you share how you're using this prop in your app(s)?

bySabi:
I use throttle on apps that animated scroll to hash links.
User generated scroll events rarely will have any performance impact but programmatic generated events could degrade it mostly on mobile environments.
I´m in favor of remove it too but, IMO, we must wait until `passive events` goes mainstrem on major browsers and hardware constrained devices.
One thing to take in consideration is that `throtteHandler` prop actually not throttled anything is just a Doc convention. Maybe the prop name must be change in order of avoid any confusion but retain it like a `_handleScroll` bridge is useful in some cases, `throttling` will be one.

And finally I will like to say that I don´t like the current direction of React-waypoint. Latest addition: throttleHandler, debug, bottom/top offset, TargetEventHandler and for coming had become React-waypoint too bloated. Still think that this project must be split on a scroll metal/DOM/debug/performance part and a Pure React Component like I propose some times ago.

trotzig:
Thanks for that bit of input @bySabi. It's interesting about the programmatically powered scroll. 

@lencioni: do you have any idea when passive scroll listeners might be implemented in other browsers? 

I agree about recent props adding a bit of bloat, so removing `throttleHandler` should be a good thing. The `debug` prop too shouldn't be necessary, though it has helped me out when responding to support requests a few times. The `bottomOffset` `topOffset` made the tool a lot more powerful. Those props are a little bit harder to use than the old `threshold`, but I can't think of a good way to simplify. 

[...]

PR #142, opened 2016-12-12
Remove support for throttleHandler prop

trotzig:
This prop has had people confused. It was added to allow people to
better control scrolling performance in their applications. Since we
added it, we've done other changes to improve performance (3ad940e,
c67112) which has made the benefit of throttling questionable. Removing
it will reduce the complexity of the component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## master (unreleased)
 
+- Remove `throttleHandler`
+
 ## 4.1.0
 
 - Add `horizontal` prop. Use it to make the waypoint trigger on horizontal scrolling.
--- a/README.md
+++ b/README.md
@@ -151,13 +151,6 @@ below) has changed.
      * things down significantly, so it should only be used during development.
      */
     debug: PropTypes.bool,
-
-    /**
-     * The `throttleHandler` prop provides a function that throttle the internal
-     * scroll handler to increase performance.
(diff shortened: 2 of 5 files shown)


lencioni:
#129

trotzig:
@lencioni: any ideas why we're seeing this failure now?
 
```
Oops! Something went wrong! :(
ESLint couldn't find the plugin "eslint-plugin-import". This can happen for a couple different reasons:
```

I have a vague memory of seeing this happen unexpectedly in the past, but I can't recall from when and what the fix was.

jmeas:
@trotzig this would be the first build using `eslint-plugin-brigade@3.2.0`, which is when eslint-plugin-import was added as a dependency to that project.

An immediate "fix" to get this PR landed would be to update package.json to lock down that dep to 3.1.0. That way, this PR can land with the working version. Then, we could investigate why 3.2.0 isn't working separately.

Or we could try to figure the issue out here. I took a look at the update in 3.2.0 in that project, and nothing stood out to me as an issue. I can take another look in a bit if nobody else figures it out.

trotzig:
Ah, thanks for figuring that out @jmeas! I don't see anything wrong with 3.2.0 of eslint-config-brigade as well. It correctly specifies `eslint-plugin-import` as a dependency. @steverice: any ideas here? 

I'll go ahead and lock the version to 3.1.0 in the meantime.

jmeas:
lgtm :+1:

PR #146, opened 2017-01-04
add typescript types definition file

nicolas-schmitt:
Hi there, this pull request adds the typescript 2 type definitions to your module.
I used your documentation to create the file and followed [those guidelines](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

I'm just wondering about wether or not have a dependency on @types/react.

Let me know if you need anything fixed :smile_cat:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="react" />
 
-import {ClassicComponentClass} from 'react';
+import { ClassicComponentClass } from 'react';
 
 declare namespace ReactWaypoint {
     interface CallbackArgs {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,111 @@
+/// <reference types="react" />
+
+import {ClassicComponentClass} from 'react';
+
+declare namespace ReactWaypoint {
+    interface CallbackArgs {
+        /*
+         * The position that the waypoint has at the moment.
+         * One of Waypoint.below, Waypoint.above, Waypoint.inside, and Waypoint.invisible.
+         */
+        currentPosition: string;
+
+        /*
+         * The position that the waypoint had before.
+         * One of Waypoint.below, Waypoint.above, Waypoint.inside, and Waypoint.invisible.
+         */
+        previousPosition: string;
+
+        /*
+         * The native scroll event that triggered the callback.
+         * May be missing if the callback wasn't triggered as the result of a scroll
+         */
+        event?: Event;
+
+        /*
+         * The waypoint's distance to the top of the viewport.
+         */
+        waypointTop: number;
+
+        /*
+         * The distance from the scrollable ancestor to the viewport top.
+         */
+        viewportTop: number;
+
+        /*
+         * The distance from the bottom of the scrollable ancestor to the viewport top.
+         */
+        viewportBottom: number;
+    }
+
+    interface WaypointProps {
+        /**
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,111 @@
+/// <reference types="react" />
+
+import {ClassicComponentClass} from 'react';
```
trotzig: nit: spaces inside destructured assignment: 

```js
import { ClassicComponentClass } from 'react';
```

nicolas-schmitt: It does help editors indeed but it's also required by the typescript transpiler.

Without a type definition, the typescript transpiler will throw error such as these :
> ERROR in ./app/components/my-component.tsx
> (136,17): error TS2604: JSX element type 'Waypoint' does not have any construct or call signatures.
> 
> ERROR in ./app/components/my-component.tsx
> (10,28): error TS2307: Cannot find module 'react-waypoint'.
Depending on your configuration, it might break your build. At the very least it clutters the output with meaningless errors.

As stated in the publishing guide, it can be bundled with the module or live a in a separate one (most likely @types/react-waypoint). The former makes it simpler to use since you don't have to look for a definition file manually.

You can ask the transpiler to generate such file but for that you'll obviously have to code in typescript which I'm guessing isn't what you're aiming for :)

nicolas-schmitt:
It does help editors indeed but it's also required by the typescript transpiler.

Without a type definition, the typescript transpiler will throw error such as these :
> ERROR in ./app/components/my-component.tsx
> (136,17): error TS2604: JSX element type 'Waypoint' does not have any construct or call signatures.
> 
> ERROR in ./app/components/my-component.tsx
> (10,28): error TS2307: Cannot find module 'react-waypoint'.
Depending on your configuration, it might break your build. At the very least it clutters the output with meaningless errors.

As stated in the publishing guide, it can be bundled with the module or live a in a separate one (most likely @types/react-waypoint). The former makes it simpler to use since you don't have to look for a definition file manually.

You can ask the transpiler to generate such file but for that you'll obviously have to code in typescript which I'm guessing isn't what you're aiming for :)

trotzig:
Cool, thanks for explaining. 

As for keeping the definitions in sync, are we left with having to check manually?

nicolas-schmitt:
Sadly yes. That's what's happening on all other modules I know. 
They get updated fairly quickly though since an outdated definition might break your build.

trotzig:
Is there a cheap way for us to add a build-step catching out-of-sync issues?

nicolas-schmitt:
It depends on what you call cheap.
What can be done is writing some kind of unit tests but instead of an assertion library like jasmine or mocha you would use the typescript transpiler.
The idea is this: if it builds, then your definitions are ok. So basically, you write simple implementations of waypoint an try to build them.

like,
`<Waypoint topOffset="20%">`
will work

but
`<Waypoint topOffffffset="20%">`
will throw an unknown property error.

this would require to:
- install typescript as a dev dependency,
- write the tests
- maintain the tests
- add a simple npm command to run them

As far as i know, that's the only way.

trotzig:
Thanks for explaining things to me. I'm okay merging this as-is. If we notice that drift becomes an issue we can always add a test later on.

trotzig:
I just released 5.0.0 including this change.

nicolas-schmitt:
much appreciated :smile_cat:

PR #147, opened 2017-01-06
Update type definition to support ES6 default imports

fongandrew:
@nicolas-schmitt, this type definition assumes the module is a CommonJS/AMD module rather than an ES6 one, which forces us to import the module like so if we're using TypeScript:

`import Waypoint = require('react-waypoint')`

But react-waypoint seems like it's an ES6 module with a default export, so in theory, this should work:

`import Waypoint from 'react-waypoint'`

However, the type definition would need to indicate this. By adding a `default` attribute to the type definition, this would allow both styles of imports to work.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -108,4 +108,4 @@ declare namespace ReactWaypoint {
 }
 
 declare let Waypoint: ReactWaypoint.Waypoint;
-export = Waypoint;
+export default Waypoint;


lencioni:
@nicolas-schmitt will you review this please?

jmeas:
@nicolas-schmitt , clarifying question: do you mean updating the transpiler to leave the ES2015 modules intact in the distributed code? Or do you mean splitting out the current (pre-transpiled) statement of `export default class Waypoint` into, say, something like:

```js
const Waypoint = ...;

export default Waypoint;
```

and then letting that get transpiled so that the ES2015 modules go away?

Or something else? Sorry if this question seems super duper basic, I've never used TypeScript!

nicolas-schmitt:
@jmeas What i meant is : @fongandrew wishes to import Waypoint as an ES6 module, which it's good since Waypoint actually is one.

When i submited my definitions i overlooked that fact and the current version forces typescript users to either import it this way:
`import Waypoint = require('react-waypoint');`
or that way:
`import * as Waypoint from 'react-waypoint';`

To fix that he suggested to add a "default" property to the Waypoint interface (line 107).
It works but I would rather change the export statement in the d.ts instead (line 111 before his change).

IE: replacing
`export = Waypoint;`
by
`export default Waypoint;`

Which would be more accurate.

does that make sense ?

jmeas:
Yes, that does make sense! Thank you for the clarification.

fongandrew:
@nicolas-schmitt Yes, that's the right way to do it. My concern is that this would force TypeScript users importing the module as a CommonJS module (i.e. `require('react-waypoint')` -- which is what the README suggests) to do `require('react-waypoint').default` instead.

Basically the question is whether the type definition should accurately model the source code (ES6) or the Babel transpilation (which because of the `add-module-exports` plugin allows both `require('react-waypoint')` and `require('react-waypoint').default`).

I think it makes sense to just model the ES6 source since ... well, it's the source and eventually all browsers will be ES6 compatible. But wanted to check that makes sense.

nicolas-schmitt:
Yes i'ld go with es6 as well since it matches the source.
Plus it would stir people into making their import the right way (using a from instead of a require statement) and typescript might eventually add a feature like add-module-exports anyway.

fongandrew:
OK, just updated commit to use `export default`.

nicolas-schmitt:
Great ! Then I guess all that's left is merging this PR :smile_cat:

trotzig:
Thanks all!

trotzig:
Released as 5.0.1

fongandrew:
Oh whoops -- can we actually revert this PR? 

I misinterpreted how the build process works. Although the `react-waypoint` source is written as an ES6 module, because of the `add-module-exports` module, Babel transforms it into a CommonJS module (and only a CommonJS module -- there is no `default` export value in the actual build). Therefore, it cannot be imported as a pure ES6 module (`import Waypoint from 'react-waypoint'` does not work -- it will be undefined).

Apologies for the inconvenience all!

PR #151, opened 2017-01-16
Clear timeout when unmounting

lencioni:
There may be some cases where waypoint mounts and then unmounts before
this timeout has an opportunity to happen. For these cases we want to
clear the timeout when the component unmounts to avoid any possible
errors.


trotzig:
👍

Issue #112, opened 2016-07-24
A way to create scrollspy


VladShcherbin:
Hey, guys. As far as I understand, waypoint creates a span element, which is checked if currently visible. This span is just a line. What if the waypoint is a component (with height say 1000px) and you want to trigger onEnter and onLeave when this component is visible.

Is it somehow possible to do?

jmeas:
So far in my apps I've used the span to generate the scrollspy (my scrollbars behave similarly to the ones in Boostrap, which only requires the single span). But I can imagine situations where one might want a scrollspy that factors in the height of some component.

One way to do this would be to add two waypoints: one at the top of your component, and one at the bottom. I know it's not ideal to do this every time, but you should be able to make a simple HoC component abstraction to abstract it. I've been toying with the idea of making a simple wrapper for my particular use case (even though I don't need height to be accounted for), so that I don't have to use the low-level API that this lib provides.

Also, there's been a lot of discussion about converting this lib itself to a HoC (see: #87 and #41 ). I admit I haven't fully read both of those issues – only skimmed them so far – but I remember it being mentioned that factoring in the height of the contents of the waypoint could be taken into account.

If this interests you, you should read up on those issues and see what's been talked about so far. Again, I haven't read them fully but from what I remember other contributors to this lib would at least consider a refactor to make this a HoC. I'm open to it, too!

Lmk if that helps :v:

jmeas:
Ah, interesting @VladShcherbin . I hadn't considered the problem of the component mounting and extending beyond both boundaries. If there was some way to determine the position of a waypoint on mount, then you could do what you're looking for in a robust way. I haven't thought too hard about doing that, but a quick glance at the API suggests that it might not be possible.

I agree that the most elegant way to handle this would be a lib that functions like your second example.

trotzig:
I too hadn't considered the mounting issue. Thanks for adding that to the discussion. If a waypoint was allowed to have children we could take that scenario into account. I'm happy to accept a PR adding support for this use-case using a regular children approach. If we can come up with a use-case that necessitates a HOC I'd be open for that too. The use-cases I've seen so far however haven't really needed a HOC approach.

bySabi:
Waypoint mounting can be delayed, Ex:

```js
class WaypointLazy extends React.Component {

  state = { render: <span style={{fontSize: 0}} /> };

  componentDidMount() {
    setTimeout(() => this.setState({ render: <Waypoint {...this.props} /> }));
  }

  render() {
[...]

PR #152, opened 2017-01-25
Accept children

rattrayalex:
Addresses #112 
Intended to supersede #41 

If you pass children to a Waypoint element, `onEnter` will be triggered when the element is fully or partially within the viewport, or if it fully encloses the viewpoint. `onLeave` will be triggered when no part of the element is within the viewport. 

This does not differentiate between full, partial, or enclosing visibility. That would likely require API changes (eg; a parameter being passed to `onEnter` and `onLeave`, new callbacks like `onPartialEnter`, or a new callback with a parameter, eg `onChange(visibility: Enum<'full'|'partial'|'enclosing'|'none'>`). Leaving as a TODO. 

I haven't added tests yet, so this isn't ready to merge. In the meantime, though, here's a screencast: 

waypoint_scroll_demo_short_720

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ below) has changed.
 />
 ```
 
-Waypoints can have children, allowing you to track when a section of content
-enters or leaves the viewport.
+Waypoints can take a child, allowing you to track when a section of content
+enters or leaves the viewport. For details, see [Children](#children), below.
 
 ```jsx
 <Waypoint onEnter={this._handleEnter}>
@@ -87,42 +87,6 @@ enters or leaves the viewport.
   </div>
 </Waypoint>
 ```
-
-Note that this inserts a `<div>` wrapping the children passed in.
-If this is undesirable, you can use the `noWrapper={true}` prop,
-in which case you can pass a single React Class or DOM element
-which will be rendered directly. You cannot pass a stateless component,
-text node, or array of nodes when `noWrapper={true}`.
-
-```jsx
-<Waypoint onEnter={this._handleEnter} noWrapper>
-  <section>
-    This section will be rendered without an enclosing div.
-  </section>
-</Waypoint>
-
-// INVALID:
-<Waypoint noWrapper>
-  <hr />
-  <section>
-    Oops, I cannot pass multiple children with noWrapper={true}!
-  </section>
-</Waypoint>
-
-// INVALID:
-<Waypoint noWrapper>
-  Oops, I must pass an element with noWrapper={true}!
-</Waypoint>
-
-// INVALID:
-<Waypoint noWrapper>
-  <MyStatelessComponent>
-    Oops, I cannot pass a stateless component with noWrapper={true}!
-    (this is because stateless components do not accept refs)
-  </MyStatelessComponent>
-</Waypoint>
-```
-
 
 ### Example: [JSFiddle Example][jsfiddle-example]
 
@@ -296,29 +260,34 @@ top boundary or the bottom boundary.
 
 ## Children
 
-If you don't pass children into your Waypoint, then you can think of the
+If you don't pass a child into your Waypoint, then you can think of the
 waypoint as a line across the page. Whenever that line crosses a
 [boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
 will be called.
 
-When children are passed, the waypoint's size will be determined by the
-size of a div wrapping the contained children (or, with `noWrapper={true}`,
-the size of the child element you pass). The `onEnter` callback will be called
-when *any* part of the children is visible in the viewport. The `onLeave`
-callback will be called when *all* children have exited the viewport.
-(Note that this is measured only on a single axis; strangely positioned elements
-may not work as expected).
+If you do pass a child, it must be a single DOM Element (eg; a `<div>`)
+and *not* a Component Element (eg; `<MyComponent />`).
 
-Deciding whether to pass children or not will depend on your use case. One
-example of when passing children is useful is for a scrollspy. Imagine if you
-want to fire a waypoint when a particularly long piece of content is visible
-onscreen. When the page loads, it is conceivable that both the top and bottom of
-this piece of content could lie outside of the boundaries, because the content
-is taller than the viewport. If you didn't pass children, and instead put the
-waypoint above or below the content, then you will not receive an `onEnter`
-callback (nor any other callback from this library). Instead, passing this long
-content as a child of the Waypoint would fire the `onEnter` callback when the
-page loads.
+The `onEnter` callback will be called when *any* part of the child is visible
+in the viewport. The `onLeave` callback will be called when *all* of the child
+has exited the viewport.
+
+(Note that this is measured only on a single axis. What this means is that for a
+Waypoint within a vertically scrolling parent, it could be off of the screen
+horizontally yet still fire an onEnter event, because it is within the vertical
+boundaries).
+
+Deciding whether to pass a child or not will depend on your use case. One
+example of when passing a child is useful is for a scrollspy
+(like [Bootstrap's](https://bootstrapdocs.com/v3.3.6/docs/javascript/#scrollspy)).
+Imagine if you want to fire a waypoint when a particularly long piece of content
+is visible onscreen. When the page loads, it is conceivable that both the top
+and bottom of this piece of content could lie outside of the boundaries,
+because the content is taller than the viewport. If you didn't pass a child,
+and instead put the waypoint above or below the content, then you will not
+receive an `onEnter` callback (nor any other callback from this library).
+Instead, passing this long content as a child of the Waypoint would fire the `onEnter`
+callback when the page loads.
 
 ## Containing elements and `scrollableAncestor`
 
--- a/README.md
+++ b/README.md
@@ -37,11 +37,11 @@ yarn add react-waypoint
 
 ## Usage
 
-```javascript
+```jsx
 var Waypoint = require('react-waypoint');
 ```
 
-```javascript
+```jsx
 <Waypoint
   onEnter={this._handleWaypointEnter}
   onLeave={this._handleWaypointLeave}
@@ -60,7 +60,7 @@ position. Sometimes it's useful to have a waypoint that fires `onEnter` every
 time it is updated as long as it stays visible (e.g. for infinite scroll). You
 can then use a `key` prop to control when a waypoint is reused vs. re-created.
 
-```javascript
+```jsx
 <Waypoint
   key={cursor}
   onEnter={this._loadMoreContent}
@@ -71,7 +71,7 @@ Alternatively, you can also use an `onPositionChange` event to just get
 notified when the waypoint's position (e.g. inside the viewport, above or
 below) has changed.
 
-```javascript
+```jsx
 <Waypoint
   onPositionChange={this._handlePositionChange}
 />
@@ -80,13 +80,46 @@ below) has changed.
 Waypoints can have children, allowing you to track when a section of content
 enters or leaves the viewport.
 
-```javascript
-<Waypoint
-  onPositionChange={this._handlePositionChange}
->
+```jsx
+<Waypoint onEnter={this._handleEnter}>
   <div>
     Some content here
   </div>
+</Waypoint>
+```
+
+Note that this inserts a `<div>` wrapping the children passed in.
+If this is undesirable, you can use the `noWrapper={true}` prop,
+in which case you can pass a single React Class or DOM element
+which will be rendered directly. You cannot pass a stateless component,
+text node, or array of nodes when `noWrapper={true}`.
+
+```jsx
+<Waypoint onEnter={this._handleEnter} noWrapper>
+  <section>
+    This section will be rendered without an enclosing div.
+  </section>
+</Waypoint>
+
+// INVALID:
+<Waypoint noWrapper>
+  <hr />
+  <section>
+    Oops, I cannot pass multiple children with noWrapper={true}!
+  </section>
+</Waypoint>
+
+// INVALID:
+<Waypoint noWrapper>
+  Oops, I must pass an element with noWrapper={true}!
+</Waypoint>
+
+// INVALID:
+<Waypoint noWrapper>
+  <MyStatelessComponent>
+    Oops, I cannot pass a stateless component with noWrapper={true}!
+    (this is because stateless components do not accept refs)
+  </MyStatelessComponent>
 </Waypoint>
 ```
 
@@ -97,7 +130,7 @@ enters or leaves the viewport.
 
 ## Prop types
 
-```javascript
+```jsx
   propTypes: {
 
     /**
@@ -268,10 +301,13 @@ waypoint as a line across the page. Whenever that line crosses a
 [boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
 will be called.
 
-When children are passed, then the waypoint's size will be determined by the
-size of the contained children. The `onEnter` callback will be called when *any*
-part of the children is visible in the viewport. The `onLeave` callback will be
-called when *all* children have exited the viewport.
+When children are passed, the waypoint's size will be determined by the
+size of a div wrapping the contained children (or, with `noWrapper={true}`,
+the size of the child element you pass). The `onEnter` callback will be called
+when *any* part of the children is visible in the viewport. The `onLeave`
+callback will be called when *all* children have exited the viewport.
+(Note that this is measured only on a single axis; strangely positioned elements
+may not work as expected).
 
 Deciding whether to pass children or not will depend on your use case. One
 example of when passing children is useful is for a scrollspy. Imagine if you
--- a/README.md
+++ b/README.md
@@ -77,6 +77,19 @@ below) has changed.
 />
 ```
 
+Waypoints can have children, allowing you to track when a section of content
+enters or leaves the viewport.
+
+```javascript
+<Waypoint
+  onPositionChange={this._handlePositionChange}
+>
+  <div>
+    Some content here
+  </div>
+</Waypoint>
+```
+
 
 ### Example: [JSFiddle Example][jsfiddle-example]
 
@@ -217,7 +230,7 @@ then the boundaries will be pushed inward, toward the center of the page. If
 you specify a negative value for an offset, then the boundary will be pushed
 outward from the center of the page.
 
-#### Horizontal Scrolling
+#### Horizontal Scrolling Offsets and Boundaries
 
 By default, waypoints listen to vertical scrolling. If you want to switch to
 horizontal scrolling instead, use the `horizontal` prop. For simplicity's sake,
@@ -248,6 +261,29 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
 passed to the callbacks, you can determine whether the waypoint has crossed the
 top boundary or the bottom boundary.
 
+## Children
+
+If you don't pass children into your Waypoint, then you can think of the
+waypoint as a line across the page. Whenever that line crosses a
+[boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
+will be called.
+
+When children are passed, then the waypoint's size will be determined by the
+size of the contained children. The `onEnter` callback will be called when *any*
+part of the children is visible in the viewport. The `onLeave` callback will be
+called when *all* children have exited the viewport.
+
+Deciding whether to pass children or not will depend on your use case. One
+example of when passing children is useful is for a scrollspy. Imagine if you
+want to fire a waypoint when a particularly long piece of content is visible
+onscreen. When the page loads, it is conceivable that both the top and bottom of
+this piece of content could lie outside of the boundaries, because the content
+is taller than the viewport. If you didn't pass children, and instead put the
+waypoint above or below the content, then you will not receive an `onEnter`
+callback (nor any other callback from this library). Instead, passing this long
+content as a child of the Waypoint would fire the `onEnter` callback when the
+page loads.
+
 ## Containing elements and `scrollableAncestor`
 
 React Waypoint positions its [boundaries](#offsets-and-boundaries) based on the
@@ -277,6 +313,7 @@ This might look something like:
 ```
 
 ## Troubleshooting
+
 If your waypoint isn't working the way you expect it to, there are a few ways
 you can debug your setup.
 
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -34,8 +34,21 @@ function getCurrentPosition(bounds) {
     return POSITIONS.invisible;
   }
 
+  // top is within the screen
   if (bounds.viewportTop <= bounds.waypointTop &&
       bounds.waypointTop <= bounds.viewportBottom) {
+    return POSITIONS.inside;
+  }
+
+  // bottom is within the screen
+  if (bounds.viewportTop <= bounds.waypointBottom &&
+      bounds.waypointBottom <= bounds.viewportBottom) {
+    return POSITIONS.inside;
+  }
+
+  // top is above the screen and bottom is below the screen
+  if (bounds.waypointTop <= bounds.viewportTop &&
+      bounds.viewportBottom <= bounds.waypointBottom) {
     return POSITIONS.inside;
   }
 
@@ -255,6 +268,7 @@ export default class Waypoint extends React.Component {
       previousPosition,
       event,
       waypointTop: bounds.waypointTop,
+      waypointBottom: bounds.waypointBottom,
       viewportTop: bounds.viewportTop,
       viewportBottom: bounds.viewportBottom,
     };
@@ -279,6 +293,7 @@ export default class Waypoint extends React.Component {
         previousPosition,
         event,
         waypointTop: bounds.waypointTop,
+        waypointBottom: bounds.waypointBottom,
         viewportTop: bounds.viewportTop,
         viewportBottom: bounds.viewportBottom,
       });
@@ -287,6 +302,7 @@ export default class Waypoint extends React.Component {
         previousPosition: POSITIONS.inside,
         event,
         waypointTop: bounds.waypointTop,
+        waypointBottom: bounds.waypointBottom,
         viewportTop: bounds.viewportTop,
         viewportBottom: bounds.viewportBottom,
       });
@@ -295,8 +311,9 @@ export default class Waypoint extends React.Component {
 
   _getBounds() {
     const horizontal = this.props.horizontal;
-    const waypointTop = horizontal ? this._ref.getBoundingClientRect().left :
-      this._ref.getBoundingClientRect().top;
+    const { left, top, right, bottom } = this._ref.getBoundingClientRect();
+    const waypointTop = horizontal ? left : top;
+    const waypointBottom = horizontal ? right : bottom;
 
     let contextHeight;
     let contextScrollTop;
@@ -313,6 +330,7 @@ export default class Waypoint extends React.Component {
 
     if (this.props.debug) {
       debugLog('waypoint top', waypointTop);
+      debugLog('waypoint bottom', waypointBottom);
       debugLog('scrollableAncestor height', contextHeight);
       debugLog('scrollableAncestor scrollTop', contextScrollTop);
     }
@@ -324,6 +342,7 @@ export default class Waypoint extends React.Component {
 
     return {
       waypointTop,
+      waypointBottom,
       viewportTop: contextScrollTop + topOffsetPx,
       viewportBottom: contextBottom - bottomOffsetPx,
     };
@@ -333,6 +352,10 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
+    if (this.props.children) {
+      return <div ref={this.refElement}>{this.props.children}</div>;
+    }
+
     // We need an element that we can locate in the DOM to determine where it is
     // rendered relative to the top of its context.
     return <span ref={this.refElement} style={{ fontSize: 0 }} />;
@@ -340,6 +363,7 @@ export default class Waypoint extends React.Component {
 }
 
 Waypoint.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.array]),
   debug: PropTypes.bool,
   onEnter: PropTypes.func,
   onLeave: PropTypes.func,
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -34,19 +34,19 @@ function getCurrentPosition(bounds) {
     return POSITIONS.invisible;
   }
 
-  // top is within the screen
+  // top is within the viewport
   if (bounds.viewportTop <= bounds.waypointTop &&
       bounds.waypointTop <= bounds.viewportBottom) {
     return POSITIONS.inside;
   }
 
-  // bottom is within the screen
+  // bottom is within the viewport
   if (bounds.viewportTop <= bounds.waypointBottom &&
       bounds.waypointBottom <= bounds.viewportBottom) {
     return POSITIONS.inside;
   }
 
-  // top is above the screen and bottom is below the screen
+  // top is above the viewport and bottom is below the viewport
   if (bounds.waypointTop <= bounds.viewportTop &&
       bounds.viewportBottom <= bounds.waypointBottom) {
     return POSITIONS.inside;
@@ -352,6 +352,17 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
+    if (this.props.noWrapper) {
+      const child = React.Children.only(this.props.children);
+      const ref = (node) => {
+        this.refElement(node);
+        if (this.props.children.ref) {
+          this.props.children.ref(node);
+        }
+      };
+      return React.cloneElement(child, { ref });
+    }
+
     if (this.props.children) {
       return <div ref={this.refElement}>{this.props.children}</div>;
     }
@@ -363,7 +374,8 @@ export default class Waypoint extends React.Component {
 }
 
 Waypoint.propTypes = {
-  children: PropTypes.oneOfType([PropTypes.element, PropTypes.array]),
+  children: PropTypes.node,
+  noWrapper: PropTypes.bool,
   debug: PropTypes.bool,
   onEnter: PropTypes.func,
   onLeave: PropTypes.func,
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -120,6 +120,35 @@ function computeOffsetPixels(offset, contextHeight) {
   }
 }
 
+/**
+ * When an element's type is a string, it represents a DOM node with that tag name
+ * https://facebook.github.io/react/blog/2015/12/18/react-components-elements-and-instances.html#dom-elements
+ *
+ * @param {React.element} Component
+ * @return {bool} Whether the component is a DOM Element
+ */
+function isDOMElement(Component) {
+  return (typeof Component.type === 'string');
+}
+
+
+/**
+ * Raise an error if "children" isn't a single DOM Element
+ *
+ * @param {React.element|null} children
+ * @return {undefined}
+ */
+function ensureChildrenIsSingleDOMElement(children) {
+  if (children) {
+    React.Children.only(children);
+
+    if (!isDOMElement(children)) {
+      throw new Error(
+        'You must wrap any Component Elements passed to Waypoint in a DOM Element (eg; a <div>).'
+      );
+    }
+  }
+}
 
 /**
  * Calls a function when you scroll to the element.
@@ -132,6 +161,8 @@ export default class Waypoint extends React.Component {
   }
 
   componentWillMount() {
+    ensureChildrenIsSingleDOMElement(this.props.children);
+
     if (this.props.scrollableParent) { // eslint-disable-line react/prop-types
       throw new Error('The `scrollableParent` prop has changed name to `scrollableAncestor`.');
     }
@@ -169,6 +200,10 @@ export default class Waypoint extends React.Component {
     this.initialTimeout = setTimeout(() => {
       this._handleScroll(null);
     }, 0);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    ensureChildrenIsSingleDOMElement(nextProps.children);
   }
 
   componentDidUpdate() {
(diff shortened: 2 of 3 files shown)


rattrayalex:
@jmeas I could use some help/direction with tests. I copy/pasted several of the main tests that seemed like they could use duplication with children provided, but my hunch is this is largely a waste. 

Instead, one or two tests ensuring that fully-enclosing and top-above, bottom-visible work should be sufficient, but I struggled to write those effectively. Any chance you could sketch out a skeleton for those?

src/waypoint.jsx
```diff
@@ -333,13 +352,20 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
-    // We need an element that we can locate in the DOM to determine where it is
-    // rendered relative to the top of its context.
-    return <span ref={this.refElement} style={{ fontSize: 0 }} />;
+    if (this.props.children) {
+      // TODO: consider returning children directly when there is only one child,
+      // using React.cloneElement to attach a ref (but child might already have a ref)
```
trotzig: I think the wrapping div might be just fine, though I understand it if we want to limit the number of DOM nodes.

src/waypoint.jsx
```diff
@@ -295,8 +311,9 @@ export default class Waypoint extends React.Component {
 
   _getBounds() {
     const horizontal = this.props.horizontal;
-    const waypointTop = horizontal ? this._ref.getBoundingClientRect().left :
-      this._ref.getBoundingClientRect().top;
+    const rect = this._ref.getBoundingClientRect();
```
trotzig: nit: you could deconstruct this:

```js
const { left, top, right, bottom } = this._ref.getBoundingClientRect();
```

src/waypoint.jsx
```diff
@@ -34,11 +34,27 @@ function getCurrentPosition(bounds) {
...
   }
 
+  // bottom is within the screen
+  if (bounds.viewportTop <= bounds.waypointBottom &&
+      bounds.waypointBottom <= bounds.viewportBottom) {
+    return POSITIONS.inside;
+  }
+
+  // TODO: consider returning more granular values than 'inside',
+  // eg; `partiallyInside`, `fullyInside`, and `fullyEnclosing`
```
trotzig: I don't know your exact use-case, but what you have seems perfectly fine to start with.

trotzig:
I noticed you found a way to get the specs working. I agree that they're not easy to use/update, I'm to blame for that. There's too much indirection, and I think the tests could benefit from a refactor where we would break it up into smaller pieces, and preferably avoided most of the indirection. 

That being said, I hacked together an example of tests I think would be sufficient for this change: https://gist.github.com/trotzig/4321f7846c35039563411b94498a7d5c

The first test is nested under `when the Waypoint is below the bottom`. The second one couldn't be nested so I duplicated some stuff. Feel free to use this or just stick with what you have. 

Thanks for contributing!

src/waypoint.jsx
rattrayalex: It's also a tad unfortunate to have to inject an element which may interfere with syling. But I can remove the TODO because it's unlikely to be worth it

rattrayalex:
Awesome, thanks for the starting point @trotzig ! I'll get cracking on that shortly.

rattrayalex:
Awesome, update! @trotzig let me know if this looks good and I'll squash the commits

src/waypoint.jsx
jmeas: I understand the desire to keep the number of nodes small. I wonder if returning `children` directly would require a refactoring of `_getBounds` to account for an array of children, as we wouldn't be able to rely on a single element's `getBoundingClientRect`?

In any event, I agree with @trotzig that this is good as-is :+1:

rattrayalex: 😄 awesome!

@jmeas I don't think that's allowed in any case – https://github.com/facebook/react/issues/2127. IIRC, `children` is either an array or a single element (not wrapped in an array), so when it's a single element, you can return it directly (after adding a `ref` attribute). But that's a lot of work for a potentially limited usecase

jmeas: You're right, @rattrayalex  – it's not currently possible. I'm thinking about the future when [Fiber](https://github.com/acdlite/react-fiber-architecture) supports it. The current code prepares us better for that day.

rattrayalex: Wow, looks like Fiber may be close to shipping! Exciting!

src/waypoint.jsx
```diff
@@ -34,11 +34,24 @@ function getCurrentPosition(bounds) {
...
     return POSITIONS.inside;
   }
 
+  // bottom is within the screen
+  if (bounds.viewportTop <= bounds.waypointBottom &&
+      bounds.waypointBottom <= bounds.viewportBottom) {
+    return POSITIONS.inside;
+  }
+
+  // top is above the screen and bottom is below the screen
```
jmeas: I think this issue goes beyond this PR, but we should be consistent about the name that we give the "containing element that is scrollable." Here it's being called "screen," the prop to manually set it it is "scrollableAncestor," and when the offsets are taken into account, its called the "viewport." Does anyone else think this could affect someone's understanding of the code? If not, then maybe it's no big deal.

src/waypoint.jsx
```diff
@@ -34,11 +34,24 @@ function getCurrentPosition(bounds) {
...
+  if (bounds.waypointTop <= bounds.viewportTop &&
+      bounds.viewportBottom <= bounds.waypointBottom) {
```
jmeas: Doesn't matter too much, but we could get away with these just be `<`. The `==` case should be handled by the above two conditionals.

src/waypoint.jsx
```diff
@@ -34,11 +34,24 @@ function getCurrentPosition(bounds) {
...
+    return POSITIONS.inside;
+  }
+
   if (bounds.viewportBottom < bounds.waypointTop) {
     return POSITIONS.below;
   }
```
jmeas: It may make sense to make the "outside" conditionals first. I suspect on a page with lots of waypoints, most of them will not be "inside," so this would allow us to skip over the other conditions. This may just be a micro-optimization with no real performance gain, though. What do y'all think?

jmeas:
This is really good. Thanks for putting this together, @rattrayalex ! I left a few comments, though none of them are blockers. They're just a few thoughts I had while reading the code.

We'll also want to update the README so that folks know they can pass children. The guides on boundaries and offsets may need to be updated, too, to account for the fact that waypoints can now have height. I can grab some of that / all of that if you want @rattrayalex . Just lmk.

Also, I haven't checked out the tests yet...about to do that now.

rattrayalex:
Thanks @jmeas ! 

Would be awesome if you could update docs, as I'm stressing my "open source" budget pretty thin for the week 😉

src/waypoint.jsx
rattrayalex: Hmm... I'm slightly concerned about the possibility of bug introduction, and the above conditionals do not seem likely to be expensive operations, so I'd lean towards leaving as-is personally.

src/waypoint.jsx
rattrayalex: Hmm... I'm slightly concerned about the possibility of bug introduction, and the above conditionals do not seem likely to be expensive operations, so I'd lean towards leaving as-is personally.

jmeas:
Sure thing. I'll update the docs today or tomorrow.

rattrayalex:
Sweet! I've squashed my commits; feel free to merge (or modify & merge) when you like!

jmeas:
Thanks for squashing @rattrayalex . I'll add an extra commit into this PR to update the docs tomorrow; didn't have 🕒 today.

README.md
```diff
@@ -248,6 +261,29 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
 
+## Children
+
+If you don't pass children into your Waypoint, then you can think of the
+waypoint as a line across the page. Whenever that line crosses a
+[boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
+will be called.
+
+When children are passed, then the waypoint's size will be determined by the
+size of the contained children. The `onEnter` callback will be called when *any*
```
lencioni: > the waypoint's size will be determined by the size of the contained children

I assume that this isn't strictly true (e.g. if the children are absolutely positioned). Perhaps we should phrase this in a way that makes this clear, or provide a list of known cases where this statement is not true.

README.md
```diff
@@ -248,6 +261,29 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+part of the children is visible in the viewport. The `onLeave` callback will be
```
lencioni: > any part of the children is visible in the viewport

This is true, per axis I believe. In other words, if you have a page that scrolls in both directions and you have a waypoint with children that is vertically in the viewport but not horizontally, it will still fire.

src/waypoint.jsx
lencioni: I agree, there is value in being consistent with terminology. It would be nice if this PR didn't make the terminology more cloudy at the least.

src/waypoint.jsx
```diff
@@ -295,8 +311,9 @@ export default class Waypoint extends React.Component {
 
   _getBounds() {
     const horizontal = this.props.horizontal;
-    const waypointTop = horizontal ? this._ref.getBoundingClientRect().left :
-      this._ref.getBoundingClientRect().top;
+    const { left, top, right, bottom } = this._ref.getBoundingClientRect();
+    const waypointTop = horizontal ? left : top;
+    const waypointBottom = horizontal ? right : bottom;
```
lencioni: top/bottom is very vertically oriented terminology. I know you were just following the local pattern here, but what do you think about using `waypointStart` and `waypointEnd` instead?

src/waypoint.jsx
```diff
@@ -333,13 +352,18 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
+    if (this.props.children) {
+      return <div ref={this.refElement}>{this.props.children}</div>;
```
lencioni: I think that adding a div here will be potentially unexpected for consumers and problematic for styling purposes. For example, if I have some content that I want to be inline-block and I wrap it in a Waypoint, how would I achieve the styling I need?

I [don't think we should allow people to pass in styles or classNames](https://brigade.engineering/don-t-pass-css-classes-between-components-e9f7ab192785) because of how much complexity that adds to people's codebases. 

Instead, we could use`React.Children.only` so the consumer is responsible for adding the wrapping element and styling it how they desire. To make this work, we will also need to add a `refToMeasure` prop, similar to what is outlined here: https://github.com/yannickcr/eslint-plugin-react/issues/678#issuecomment-232293175

What do folks think?

src/waypoint.jsx
```diff
@@ -333,13 +352,18 @@ export default class Waypoint extends React.Component {
...
+    }
+
     // We need an element that we can locate in the DOM to determine where it is
     // rendered relative to the top of its context.
     return <span ref={this.refElement} style={{ fontSize: 0 }} />;
   }
 }
 
 Waypoint.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.array]),
```
lencioni: Should this be `PropTypes.node`?

README.md
```diff
@@ -77,6 +77,19 @@ below) has changed.
 />
 ```
 
+Waypoints can have children, allowing you to track when a section of content
+enters or leaves the viewport.
+
+```javascript
```
jmeas: Cool, i'll update all of these.

README.md
jmeas: You're right. I'll add more details.

lencioni: If we move forward with [`React.Children.only`](https://github.com/brigade/react-waypoint/pull/152#discussion_r98356569), this section of documentation will need to be changed a fair amount (i.e. simplified), so it might be worth waiting to see how that pans out before spending too much time here.

README.md
jmeas: Hm, I think I get what you mean, but your example is throwing me off. If something is in the viewport, then is it not by definition within the boundaries of both axes?

README.md
jmeas: 👍Ill wait to see how things pan out.

README.md
lencioni: > If something is in the viewport, then is it not by definition within the boundaries of both axes?

Yes, but (and I could be misunderstanding the implementation here) Waypoint is only checking viewport boundaries on a single axis (either vertical or horizontal). So, Waypoint's definition of "in the viewport" is really "in the viewport on a single axis and ignoring the other axis".

src/waypoint.jsx
rattrayalex: I think that would require either a breaking change to the api (currently `waypointTop` is an exposed attribute) or inconsistent terminology between the internal codebase and the external api.

src/waypoint.jsx
rattrayalex: I think that sounds like a strong approach, personally. I would be willing to make that adjustment if desired.

src/waypoint.jsx
rattrayalex: As the PR is currently yes (good catch!). If the `React.Children.only` proposal is adopted – and I think it should be – then it should be `element`, not `node`, so that a `ref` can be injected.

src/waypoint.jsx
lencioni: Ah yes, that is a good point.

src/waypoint.jsx
jmeas: I haven't used `refToMeasure` before, and I wasn't able to figure out what that would look like from reading those links. Could someone type up a very quick example showing what that would look like for this lib?

The primary benefit of using `React.Children.only` that I see is allowing users to customize any number of props of the wrapping node. With that said, I think it's likely that most developers will just want to apply styling to the wrapper, rather than any other props, so I'm not sure how much that benefit will matter to folks.

I understand, and have experienced, many of the problems described in the "Don’t pass CSS classes between components" post, yet I disagree with the conclusion that categorically denying use of class name / style props is the best solution. I don't want to get too much into that here (although it'd be a great conversation, I'm sure!), but I figured I'd mention that I'd be fine just going the more traditional route and using class name/style props.

Anyway, I'm likely also alright with the `refToMeasure` prop as long as it's not too complex nor confusing for the consumers of this lib.

README.md
jmeas: Ah, right. I see what you mean. I'll update this and attempt to capture this behavior.

src/waypoint.jsx
rattrayalex: Yup. I'll go ahead and build that then?

I'm planning to also allow users to pass an element without a ref, and have one injected using `cloneWithProps` or equivalent, like so:

```jsx
<Waypoint onEnter={this.handleWaypointEnter}>
  <div>
    Something interesting would go here.
  </div>
</Waypoint>
```

in which `div` ends up with a `ref` created/managed by Waypoint.

lencioni: Ah true, cloning could make this a little more convenient.

If we do that, it would be good to mention in the documentation that functional components cannot be passed to Waypoint in this way, since they cannot have refs because they don't have instances. If there is an easy enough way to detect this, it might be worth making Waypoint give a helpful error message in dev.

rattrayalex: Ah, right, I'd forgotten about that. That'd be quite inconvenient/frustrating. 

Reading through [this thread](https://github.com/facebook/react/issues/4936#issuecomment-179897923) on the topic, it seems that an intermediary HOC may be the better approach. See this [example](https://github.com/ctrlplusb/react-sizeme/blob/master/src/sizeMe.js#L25). 

For example, user would still use 

```jsx
<Waypoint onEnter={this.handleWaypointEnter}>
  <div>
    Something interesting would go here.
  </div>
</Waypoint>
```

and `Waypoint`'s render method would change to something like: 

```jsx
if (this.props.children) {
  return <WaypointRefContainer ref={this.refElement}>{this.props.children}</WaypointRefContainer>;
}
return <span ref={this.refElement} style={{ fontSize: 0 }} />;
```

with `WaypointRefContainer` defined as, eg; 

```jsx
class WaypointRefContainer extends React.Component {
  render() {
    return React.Children.only(this.props.children);
  }
}
```

Thoughts?

(This approach wouldn't allow multi-element `children`, would it?)

lencioni: Ah, nice. If that works that sounds good to me. I don't think it would allow multi-element children because there would be no single containing element in the WaypointRefContainer.

rattrayalex: Bummer, turns out it doesn't work: `this._ref.getBoundingClientRect is not a function`. Looks like the `ref` attached to the HOC doesn't access the underlying DOM node after all 😕 . 

Thoughts on sticking with the original `div` implementation? While unfortunate in that it adds an extra element and could interfere with CSS, it won't break anyone's existing implementation and such issues are often easy to detect and fix. The other possibilities add a lot of mental overhead and complexity.

rattrayalex: An alternative might be to go with the `div` implementation by default, but allow the user to pass in a `refToMeasure` instead, in which case `this.props.children` could be rendered directly. 

This is a bit more complex; I'm happy to write it if y'all are okay maintaining it 😉

trotzig: I haven't followed all of the discussion around this, but I think the div alternative might be the right one at this point. It should solve most cases without too much headache. We should mention in the README that a div is added automatically, but that might be enough for this to become useful.

lencioni: I think that's fine, but I think it would still be good to roll with `React.Children.only` because it will be more painful to make it more restrictive later than it will be to make it less restrictive if we decide later that we should have gone a different route. What do you think?

halt-hammerzeit:
Looking forward to this new feature.
Good job collaborating on this you all.

src/waypoint.jsx
rattrayalex: Let's just get this "right" from the get-go; `<div>` by default, with an "advanced escape hatch" of `refToMeasure` requiring `React.Children.only` and documentation reminding that a `ref` can't attach to a stateless component. 

Makes usage easy for most people, and accommodates edge cases that are bound to occur. No reason to worry about adding/removing flexibility later.

rattrayalex:
Thanks! 

I'm planning to add the `refToMeasure` functionality discussed here: https://github.com/brigade/react-waypoint/pull/152#discussion_r99193415 . Probably tomorrow, maybe today. 

I'll also incorporate a bunch of the nits, incl docs, outlined above. @jmeas would you prefer I add extra commits on top, or squash everything down into one?

src/waypoint.jsx
rattrayalex: Unfortunately, it turned out the refToMeasure approach doesn't work, because the `ref` of the child is `null` until after the parent has rendered.

On the other hand, cloning and injecting a ref does work after all. I was afraid of overwriting a `ref`, but it was easy to preserve.

rattrayalex:
Alright, finally was able to implement a wrapper-free option. As I mentioned in an earlier comment, the original plan of `refToMeasure` didn't work because the child ref wasn't available on component mount (almost by definition). 

Would love 👀 !

src/waypoint.jsx
```diff
@@ -333,13 +352,30 @@ export default class Waypoint extends React.Component {
...
+    }
+
     // We need an element that we can locate in the DOM to determine where it is
     // rendered relative to the top of its context.
     return <span ref={this.refElement} style={{ fontSize: 0 }} />;
   }
 }
 
 Waypoint.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.array]),
```
lencioni: This should be just `PropTypes.node`

src/waypoint.jsx
```diff
@@ -333,13 +352,30 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
+    if (this.props.noWrapper) {
```
lencioni: This should only happen if also `this.props.children`

src/waypoint.jsx
```diff
@@ -333,13 +352,30 @@ export default class Waypoint extends React.Component {
...
+
     // We need an element that we can locate in the DOM to determine where it is
     // rendered relative to the top of its context.
     return <span ref={this.refElement} style={{ fontSize: 0 }} />;
   }
 }
 
 Waypoint.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.array]),
+  noWrapper: PropTypes.bool,
```
lencioni: Why provide this option? Why not make this the default? I think that would simplify the API and be clearer.

src/waypoint.jsx
rattrayalex: Yes, but someone who uses `this.props.noWrapper` without actually passing children should receive an error message – which will happen thanks to `React.Children.only`.

src/waypoint.jsx
rattrayalex: I don't think so? The following is valid: 

```jsx
<Waypoint>
  <div />
  <div />
</Waypoint>
```

It wraps the two child divs in a parent div.

src/waypoint.jsx
rattrayalex: Yeah, I struggled with that. Ultimately figured it would be confusing for too many users, especially beginners, since they would have to pass a single React Class-based Component or DOM Element. 

The "INVALID" examples I included in the docs are all perfectly reasonable things to do, and allowing them makes ReactWaypoint more approachable. 

Thoughts?

rattrayalex:
Ahh, good catch! No longer WIP. I think this is ready to go unless there are further objections / requests / nits?

src/waypoint.jsx
lencioni: `PropTypes.node` allows that. `PropTypes.array` here is redundant.

> Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types.

https://facebook.github.io/react/docs/typechecking-with-proptypes.html#react.proptypes

src/waypoint.jsx
lencioni: Ah, that's fair.

src/waypoint.jsx
lencioni: My opinion is that having multiple ways to do this will be more confusing, because it it less clear which one to choose and why you should choose one over the other. This also makes the documentation more complicated than it needs to be.

Since we have a good solution in the noWrapper branch, my preference would be to go with that only. I'd like to hear from others like @trotzig and @jmeas.

trotzig: I'm in favor of not having the `noWrapper` prop. This is tricky though. On the one hand, I think we could go with the pragmatic choice of just adding a `div` wrapper, and solve 90+% of use-cases. But then at some point the div is going to become a problem, so maybe it's worth it to address upfront. 

I think we should remove the prop and just go wrapper-less by default, unless the children being passed in is an array, a string, or a stateless component (function). Something like:

```jsx
function getContent(children) {
  if (React.Children.count(children) > 1 || isStateless(children)) {
    return <div>{children}</div>;
  } 
  return children;
}
```

This is a simplified example, we still have to assign the `ref`. Also, I don't know how hard the `isStateless` function would be to implement.

src/waypoint.jsx
rattrayalex: Ah, missed that. Thanks; amending now.

src/waypoint.jsx
lencioni: I don't think it is a good idea to conditionally add a div if there are multiple children or of the children is a functional component because that would be a pretty surprising and unexpected thing for folks. I think it would be much better to provide a useful error message and always require an only child that is not a functional component. That is clear, consistent, and flexible.

rattrayalex: The proposal from @trotzig sounds reasonable to me. My biggest concerns are: 

1) It'll be confusing that it behaves differently in different situations (but this should be solvable by looking at the docs). 
2) There may be times when you actually *want* the div inserted (though this is silly; the user could simply wrap a div themselves). 

> I think it would be much better to provide a useful error message and always require an only child that is not a functional component. That is clear, consistent, and flexible.

"Flexible" isn't quite the word I'd use for that 😉. Personally I'd be very surprised if an issue were not filed using a word more like "frustrating". 

My assumption here is that this library is intended for broad use, including by devs who don't understand React very well, and that convenience is an important consideration.  

In any case, @jmeas curious for your thoughts

lencioni: I'm all for convenience, but not when it comes at the cost of clarity

rattrayalex: In practice, the "you must pass a single class-based or dom component" restriction would end up with a recommendation of wrapping the contents of their Waypoint in a `div`. 

That's unnecessary clutter of user's code. 

Requiring the users to write their own wrapping `<div>` makes the final DOM output more clear, of course, but that's not something React users should really be thinking about anyway; it's abstracted away as components.

lencioni: > That's unnecessary clutter of user's code.

I disagree about this. It makes it very clear what is going on and puts full control in the hands of the developer. Of course, consumers could choose to abstract this away in a component that wraps Waypoint and adds a div, or we could provide a separate component that does this for folks.

```jsx
export default function WaypointDiv({ children, ...waypointProps }) {
  return <Waypoint {...waypointProps}><div>{children}</div></Waypoint>;
}
```

lencioni: sidebar: here's a related talk that I quite like http://2014.jsconf.eu/speakers/sebastian-markbage-minimal-api-surface-area-learning-patterns-instead-of-frameworks.html

jmeas: I understand the pros of cons of each argument here. I like @trotzig 's conditional-wrapping solution, in that it tries to hide the complexity to the user. That can be useful for someone just getting started, as they have less to think about. @lencioni is probably right that at other times it could cause confusion when that same person doesn't understand why the div is added sometimes, but not others. I think the user still has full control here, because the situations where Waypoint wraps would always happen when it was absolutely necessary. It simply wouldn't work otherwise.

@lencioni 's suggestion to never wrap is appealing because I like giving all of the control to the developer, even if the learning curve is slightly higher at first. I think a helpful error message could quickly help someone get up to speed with when it's necessary, and when it's not. Also, avoiding another prop is a good thing, too. But I also understand where @rattrayalex is coming from about requiring folks to put wrapping divs everywhere. That is a definite downside to this approach.

I think I'd be alright with either magically wrapping or not wrapping ever, but I'm not a fan of a `noWrapper` prop.

If I had to pick between the options, it'd be @lencioni 's suggestion to not add the wrapping prop, and require the user wrap it as needed. As for the WaypointDiv component, I'd probably prefer to not include it just to keep this library simpler, but I don't feel too strongly about it.

tl;dr: as long as we avoid `noWrapper`, I'm OK with anything else.

lencioni: > As for the WaypointDiv component, I'd probably prefer to not include it just to keep this library simpler, but I don't feel too strongly about it.

I agree. We could mention this technique in the documentation with a little code snippet if we wanted. That would likely be sufficient.

rattrayalex: Alright; I guess I'm not the one who'll have to deal with any user frustration anyway ;-) 

I'll try to put together some docs on the topic as well....

rattrayalex:
Ah; should have caught this earlier, but even passing a single class-based function doesn't really make sense, because it won't have a `clientBoundingRect()` method, even if it just returns a single react dom element.

This will actually make the docs somewhat simpler, since "you must pass a dom element, not a component" is an easier rule to understand.

rattrayalex:
Alright, I believe this to be ready for 👀 again

README.md
```diff
@@ -71,20 +71,33 @@ Alternatively, you can also use an `onPositionChange` event to just get
...
 below) has changed.
 
-```javascript
+```jsx
 <Waypoint
   onPositionChange={this._handlePositionChange}
 />
 ```
 
+Waypoints can take a child, allowing you to track when a section of content
```
lencioni: It might be a good idea to be explicit here that it must be an only child.

README.md
```diff
@@ -248,6 +261,31 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
 passed to the callbacks, you can determine whether the waypoint has crossed the
 top boundary or the bottom boundary.
 
+## Children
+
+If you don't pass children into your Waypoint, then you can think of the
```
lencioni: s/children/a child/

README.md
```diff
@@ -248,6 +261,31 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+will be called.
+
+If you do pass a child, it must be a single DOM Element (eg; a `<div>`).
+The `onEnter` callback will be called when *any* part of the child is visible
+in the viewport. The `onLeave` callback will be called when *all* of the child
+has exited the viewport (eg; both the top and bottom are either above or below).
+(Note that this is measured only on a single axis; strangely positioned elements
+may not work as expected).
+
+Deciding whether to pass children or not will depend on your use case. One
```
lencioni: Should maybe be "a child" here and lower as well?

README.md
```diff
@@ -248,6 +261,31 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+example of when passing children is useful is for a scrollspy. Imagine if you
```
lencioni: Should "scrollspy" link somewhere for folks who are unfamiliar with the term?

src/waypoint.jsx
```diff
@@ -107,6 +120,13 @@ function computeOffsetPixels(offset, contextHeight) {
   }
 }
 
+/**
+ * @param {React.element} Component
+ * @return {bool} Whether the component is a DOM Element
+ */
+function isDomElement(Component) {
```
lencioni: nit: I'd prefer for the DOM acronym to be fully capitalized here.

src/waypoint.jsx
```diff
@@ -107,6 +120,13 @@ function computeOffsetPixels(offset, contextHeight) {
...
+  return (typeof Component.type === 'string');
```
lencioni: Is this a reliable check? Is there some documentation we could link to?

src/waypoint.jsx
```diff
@@ -333,13 +359,30 @@ export default class Waypoint extends React.Component {
    * @return {Object}
    */
   render() {
+    if (this.props.children) {
+      const child = React.Children.only(this.props.children);
+      if (!isDomElement(child)) {
```
lencioni: I wonder if we should move this check into `componentWillMount` and `componentWillReceiveProps`. Not sure if throwing an error in render will cause weird issues with React. Also, using CWRP will allow us to only do this check when `children` has changed.

src/waypoint.jsx
```diff
@@ -333,13 +359,30 @@ export default class Waypoint extends React.Component {
...
+        throw new Error(
+          'You must wrap any Component Elements passed to Waypoint in a DOM Element (eg; a <div>).'
+        );
+      }
+      const ref = (node) => {
+        this.refElement(node);
+        if (this.props.children.ref) {
+          this.props.children.ref(node);
```
lencioni: nit: It might be nice to destructure `children` at the top of render.

README.md
rattrayalex: Hmm, do you think `a child` is inadequately clear?

I considered phrasing like "a single child" but was concerned that might sound like the following was illegal: 

```jsx
<Waypoint>
  <div>
    <GrandChild />
    <AnotherGrandChild />
  </div>
</Waypoint>
```

rattrayalex: It is also explicit in the `Children` section

README.md
rattrayalex: I'm not finding a concise explanation with a quick google, and the next sentence does a decent job of explaining the use-case IMO

src/waypoint.jsx
rattrayalex: > "When an element's type is a string, it represents a DOM node with that tag name" 

from [the docs](https://facebook.github.io/react/blog/2015/12/18/react-components-elements-and-instances.html#dom-elements)

I'll add to the jsdoc

README.md
jmeas: One idea would just be to link to [the Boostrap docs](http://getbootstrap.com/javascript/#scrollspy). It might seem weird, but:

1. I hadn't really heard the term before them
2. they do a good job at explaining it
3. so many people use, or have used, Bootstrap

They version their docs, so we could "lock it down" to a specific version to avoid issues if they remove the scrollspy ever.

...maybe this is a bad idea.

src/waypoint.jsx
rattrayalex: Sure, sounds good

README.md
rattrayalex: I kinda like that. I'll add it

rattrayalex:
@lencioni I believe all comments have been addressed; take another look?

README.md
```diff
@@ -248,6 +261,32 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
 top boundary or the bottom boundary.
 
+## Children
+
+If you don't pass a child into your Waypoint, then you can think of the
+waypoint as a line across the page. Whenever that line crosses a
+[boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
+will be called.
+
+If you do pass a child, it must be a single DOM Element (eg; a `<div>`).
```
jmeas: If a Component isn't passable (they're not, right?), then it might be a good idea to explicitly say that here. i.e.;

> If you do pass a child, it must be a single DOM Element (eg; a `<div>`), and **not** a Component (eg; `<MyCustomComponent>`).

README.md
```diff
@@ -248,6 +261,32 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+The `onEnter` callback will be called when *any* part of the child is visible
+in the viewport. The `onLeave` callback will be called when *all* of the child
+has exited the viewport (eg; both the top and bottom are either above or below).
+(Note that this is measured only on a single axis; strangely positioned elements
```
jmeas: "strangely positioned" is pretty vague. It's tough to word, but maybe something like:

> Note that this is measured only on a single axis. What this means is that for a Waypoint within a vertically scrolling parent, it could be off of the screen horizontally yet still fire an `onEnter` event.

or something to this effect.

README.md
rattrayalex: Nice, clarified (it had said this elsewhere, but I consolidated)

jmeas:
I just noticed how long this PR has been going on for:

image

I'm really thankful that you've kept up with it @rattrayalex . I know it can be tiring when a PR drags on forever. This is really great stuff, though, and we're so close to landing it now!

README.md
rattrayalex: Nice!

rattrayalex:
Hah, we broke 100! Thanks, @jmeas , I appreciate that 😄 I think I've reimplemented the feature at least 3 or 4 different ways by now; so it goes. Happy with where it ended up!

README.md
```diff
@@ -248,6 +258,37 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+will be called.
+
+If you do pass a child, it must be a single DOM Element (eg; a `<div>`)
+and *not* a Component Element (eg; `<MyComponent />`).
+
+The `onEnter` callback will be called when *any* part of the child is visible
+in the viewport. The `onLeave` callback will be called when *all* of the child
+has exited the viewport.
+
+(Note that this is measured only on a single axis.
```
jmeas: smallest teeniest ittiest nit ever...pretty early line break here :)

README.md
```diff
@@ -248,6 +258,37 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+What this means is that for a Waypoint within a vertically scrolling parent,
+it could be off of the screen horizontally yet still fire an onEnter event,
+because it is within the vertical boundaries.).
```
jmeas: > .).

seems like weird punctuation

README.md
lencioni: Yeah what you have is probably fine here. If there's confusion, we can adjust the docs then.

src/waypoint.jsx
```diff
@@ -178,6 +207,17 @@ export default class Waypoint extends React.Component {
     clearTimeout(this.initialTimeout);
   }
 
+  _ensureChildrenIsSingleDOMElement(children) {
```
lencioni: Since this function doesn't reference `this`, it doesn't need to be on the class. Move it up to the module.

lencioni:
One small change and then I think this is ready to go. I'd like to hear from @trotzig on where we landed before merging though.

src/waypoint.jsx
rattrayalex: done

src/waypoint.jsx
```diff
@@ -158,6 +202,10 @@ export default class Waypoint extends React.Component {
     }, 0);
   }
 
+  componentWillReceiveProps(nextProps) {
+    ensureChildrenIsSingleDOMElement(nextProps.children);
```
lencioni: I wonder if it would be worth only calling this if `this.props.children !== nextProps.children`

rattrayalex: I don't think there would be a downside to that...

README.md
```diff
@@ -248,6 +258,37 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+waypoint as a line across the page. Whenever that line crosses a
+[boundary](#offsets-and-boundaries), then the `onEnter` or `onLeave` callbacks
+will be called.
+
+If you do pass a child, it must be a single DOM Element (eg; a `<div>`)
+and *not* a Component Element (eg; `<MyComponent />`).
+
+The `onEnter` callback will be called when *any* part of the child is visible
+in the viewport. The `onLeave` callback will be called when *all* of the child
+has exited the viewport.
```
trotzig: I like this section.

README.md
```diff
@@ -248,6 +258,37 @@ the `onLeave` and `onEnter` callback will be called. By using the arguments
...
+example of when passing a child is useful is for a scrollspy
+(like [Bootstrap's](https://bootstrapdocs.com/v3.3.6/docs/javascript/#scrollspy)).
+Imagine if you want to fire a waypoint when a particularly long piece of content
+is visible onscreen. When the page loads, it is conceivable that both the top
+and bottom of this piece of content could lie outside of the boundaries,
+because the content is taller than the viewport. If you didn't pass a child,
+and instead put the waypoint above or below the content, then you will not
+receive an `onEnter` callback (nor any other callback from this library).
+Instead, passing this long content as a child of the Waypoint would fire the `onEnter`
+callback when the page loads.
```
trotzig: This is a good piece of context.

jmeas:
Woo!! 🎉 👯 💃 🍰 🎂

rattrayalex:
Woohoo! Thanks so much everyone!

PR #154, opened 2017-02-02
Update to allow for dynamic offset values

brandondurham:
I needed the capability to change the offset value based on some factors that Waypoint couldn’t (and shouldn’t) be aware of. The easiest way to do this was to use a function for `topOffset` that calculates the appropriate offset based on current app state:

```
topOffset={ () => this.state.licenseBarInView ? HCo.licenseBar.enterOffset : HCo.licenseBar.leaveOffset }
```

In my case that translates to `topOffset={ () => this.state.licenseBarInView ? 20 : 104 }`

Let me know if you think this is a viable change, and if I’ve missed anything here!

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ below) has changed.
     topOffset: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number,
+      PropTypes.func,
     ]),
 
     /**
@@ -127,6 +128,7 @@ below) has changed.
     bottomOffset: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number,
+      PropTypes.func,
     ]),
 
     /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -64,12 +64,12 @@ declare namespace ReactWaypoint {
 
         /**
          * `topOffset` can either be a number, in which case its a distance from the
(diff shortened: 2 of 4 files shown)


lencioni:
Does this not work?

```jsx
topOffset={this.state.licenseBarInView ? 20 : 104}
```

brandondurham:
My example was unsuitable.

I won't always be working with static values. There is in fact an occasion where the offset isn't known until an element is rendered and later measured.

jmeas:
@brandondurham , I'm not sure I understand why this change is needed. If an element is rendered, and later measured, then you could update state at that time to reflect the new measured value. Then, the static value would work the same as this callback.

If I understand this PR correctly, it's not allowing users to do something with Waypoint that they can't already do. Is that right, or am I missing something?

jmeas:
@brandondurham , any chance you've got a minute to elaborate more on this PR?

Thanks!

brandondurham:
Heya @jmeas — I was in fact able to solve for my issue without this change. I can’t in fact conjure up a scenario where this would be needed, so you can squash it.

Thanks for looking at it!

jmeas:
@brandondurham , that's awesome! I'm glad to hear you were able to solve your issue!

PR #162, opened 2017-03-08
Update devDependencies

lencioni:


jmeas:
Huh, so the failure is due to eslint-plugin-imports not being found. It is being installed, but it's nested within the brigade lint rules folder within node_modules, which is unexpected...I wonder if it's somehow because of the (small) difference between this module's eslint version and the brigade lint rules' version. But they _are_ compatible. I'm not too sure.

I'm more used to ESLint config libraries using peerDeps than deps, but with my understanding of the behavior of npm3+, I'd expect regular deps to work the same as long as the versions line up. Which they do in this situation.

As an aside, I usually like to see ^ minor and bug fix upgrades going in the opposite direction: being more open, to take full advantage of semver. i.e.; going from `^2.3.0` to `^2.0.0`. Using ^ **and** locking down the versions to be slightly contradictory to me. It's not a huge deal, but I'm curious to hear what you think the benefit of restricting the range of the ^ deps is.

lencioni:
@MatthewHerbst is this CI failure something you can look into? It seems like there might be a problem with the way eslint-config-brigade specifies its dependencies. It might be worth looking at other eslint config packages like [eslint-config-airbnb-base](https://github.com/airbnb/javascript/blob/772bbb5b7d2f6990e519c3d70539f807257492fe/packages/eslint-config-airbnb-base/package.json#L48), which use a combination of devDependencies and peerDependencies.

> As an aside, I usually like to see ^ minor and bug fix upgrades going in the opposite direction: being more open, to take full advantage of semver. i.e.; going from ^2.3.0 to ^2.0.0. Using ^ and locking down the versions to be slightly contradictory to me. It's not a huge deal, but I'm curious to hear what you think the benefit of restricting the range of the ^ deps is.

I see the benefit as making it so people who develop on this project will have more consistent versions, which should result in less unexpected behavior across machines e.g. in CI.

MatthewHerbst:
@lencioni I'll look into it now. I haven't bumped the version of `eslint-config-brigade` (or pushed the new version to npm) yet so I wouldn't expect to see failures due to recent changes unless someone is pulling directly from master. I'll update soon.

MatthewHerbst:
It seems that we failed to tag `eslint-config-brigade` v3.2.0 when we made it, so this likely _is_ pulling from master.

MatthewHerbst:
@lencioni I've created a tagged version for v3.2.0 in `eslint-config-brigade` that corresponds with when we release v3.2.0 back in December. Can you try rerunning the build jobs?

lencioni:
@MatthewHerbst ok, I restarted the builds. Looks like same problem. I don't think github releases have anything to do with npm package releases, so I don't think you changed anything by adding the tag.

MatthewHerbst:
It's an npm issue: https://travis-ci.org/brigade/react-waypoint/builds/209102287
Switching to npm3 the tests pass on node 5/6.

<3 JavaScript. I'll keep looking at this after lunch.

MatthewHerbst:
Alright:

tl;dr: The issue is directly caused by [this change/line in `eslint-config-brigade`](https://github.com/brigade/eslint-config-brigade/pull/12/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R24). It seems that because `eslint-plugin-import` is specified as a `dependency` rather than as a `devDependency`, npm2 is failing to install it properly for an unknown reason.

The issue is specific to npm2. The issues does not present when using npm >= 3, or when using the `yarn` package manager.

Possible solution 1: `eslint-plugin-import` could be made a `devDependency` in `eslint-config-brigade`. I have tested doing this and it works. This could be published as a patch version 3.2.1 so that existing users would hopefully pick it up.

However, myself and some of the other engineers are Brigade strongly believe that `eslint-plugin-import` should **not** be a `devDependency`. (We do agree that `eslint-config-brigade`'s `eslint` dependency should be a `peerDependency`, and we will be making that change for the v4 release.)

We do not purport to know for what use cases our packages are used, and it's certainly possible that someone might want to use the package in a production environment, in which case if `eslint-plugin-import` was a `devDependency`, it would not get installed and the parent package would fail.

Possible solution 2: for this pr, to bump the major version of `react-waypoint` to only support npm >= 3. As part of that we would suggest adding an [`engines`](https://docs.npmjs.com/files/package.json#engines) entry into `react-waypoint`'s `package.json`.

Curious to hear your thoughts. We're willing to go with option 1 if needed, but would much, much prefer option 2.

lencioni:
eslint-plugin-import should be listed in both devDependencies and peerDependencies. See my link above to eslint-config-airbnb-base for an example of this.

MatthewHerbst:
@lencioni [npm3 does not automatically install `peerDependencies`](https://docs.npmjs.com/files/package.json#peerdependencies), but simply prints a warning if there is an issue. This does not resolve our desire for `eslint-config-brigade` to be able to work in a `NODE_ENV=production` setting and simply running `npm install`.

MatthewHerbst:
Ok, here's what we'll do:

I'll make an `eslint-config-brigade` v3.2.1 now that also installs `eslint-plugin-import` as a `peerDependency` (and keeping it as a regular `dependency` as well). 

Version 4 of `eslint-config-brigade` will **not** include `eslint-plugin-import` as a `peerDependency`, and will also specify an `engines` entry to warn people that we only officially will be supporting npm >= 3.

@lencioni just because it fixes the problem does not mean it should be done that way; we'd prefer to get to the root of the problem (npm2 not working properly) than putting a bandaid on it.

@jmeas 
> but with my understanding of the behavior of npm3+, I'd expect regular deps to work the same as long as the versions line up. Which they do in this situation.

We fully agree, which is why we were trying to stick to not adding it as a dev and peer dep, and instead suggesting that `react-waypoint` stop using the broken npm2.

MatthewHerbst:
@jmeas [this SO](http://stackoverflow.com/a/34645112/2518231
) post further explains exactly what we (and I believe you) are thinking.

To be clear, `eslint-config-brigade` does **not** meet the 3rd requirement listed.

> **When should you use peer dependencies?**
> 
> - When you are building a library to be used by other projects, and
> - This library is using some other library, and
> - You expect/need the user to work with that other library as well

If you're using `eslint-config-brigade` you're using our entire ruleset. The implementation details of that rule set should not concern the user. If you don't want to be using our entire ruleset, don't use `eslint-config-brigade`.

MatthewHerbst:
`eslint-config-brigade@3.2.1` has been released/published.

PR #163, opened 2017-03-08
Remove some unnecessary indirection in tests

lencioni:
I was thinking about making a change to waypoint, but I found the tests
to be unnecessarily difficult to follow, so I decided to first reduce
some of the indirection and nesting. I think there is more that could be
done here, but this seems like a pretty good start.

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -71,15 +71,10 @@ describe('<Waypoint>', function() {
     jasmine.clock().uninstall();
   });
 
-  describe('when called with debug=true', () => {
-    beforeEach(() => {
-      this.props.debug = true;
-    });
-
-    it('logs to the console', () => {
-      this.subject();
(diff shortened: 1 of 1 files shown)


lencioni:
This is probably easiest reviewed with the `?w=1` query string param.

spec/waypoint_spec.js
```diff
@@ -90,9 +85,9 @@ describe('<Waypoint>', function() {
       this.scrollable = this.parentComponent;
     });
 
-    it('does not log to the console', () => {
-      expect(console.log) // eslint-disable-line no-console
-        .not.toHaveBeenCalled();
+    it('does not log to the console u', () => {
```
trotzig: did you add the "u" at the end by mistake?

lencioni: oops!

lencioni:
@trotzig Yes, that's right. I'll update the commit message

PR #166, opened 2017-03-15
Teach scrollableAncestor prop to take "window" as a string

lencioni:
When Waypoint is rendered on the server, `window` is not available.
To make Waypoint easier to work with, we allow this to be specified in
string form and safely convert to `window` here.

While I was at it, I also broke out a bunch of functions into their own
modules to make it easier to work with this project and to eventually
improve our testing story. This is probably best reviewed
commit-by-commit.

semver minor

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,9 @@ enters or leaves the viewport. For details, see [Children](#children), below.
      * the container. For example, when your target is in a div
      * that has overflow auto but you are detecting onEnter based
      * on the window.
+     *
+     * This should typically be a reference to a DOM node, but it will also work
+     * to pass it the string "window" if you are using server rendering.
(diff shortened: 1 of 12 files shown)


jmeas:
:+1:!

patrick-mcdougle:
@lencioni Just digging into this, so pardon the potentially naïve question, but why is server rendering even calling componentDidMount? Shouldn't renderToString only be calling componentWillMount?

lencioni:
@patrick-mcdougle that's a great question. I don't exactly remember the precise issue we were running into anymore, and I apparently didn't record it very well here. It may have actually been an issue in tests and not server rendering, or it may be possible that this was entirely unnecessary.

PR #169, opened 2017-03-29
Avoid unnecessary clearTimeout when unmounting

lencioni:
I was doing some profiling and noticed a pattern on pages that have a
lot of waypoints. In the timeline I can see that we spend a non-trivial
amount of time in clearTimeout (~0.25ms per waypoint on my laptop). This
can add up, so let's avoid the extra work by adding a check.

screen shot 2017-03-28 at 6 25 55 pm


trotzig:
Wow, that's surprising!

lencioni:
Yeah, I was surprised too!

lencioni:
Here's a jsperf: https://jsperf.com/cleartimeout-null-vs-int-vs-checking

PR #172, opened 2017-04-09
React 15.5

trotzig:
With React 16 soon to be released, I'm going through a process to
prepare the codebase for that version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "author": "Brigade Engineering",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^15.0.0",
+    "react": "^0.14.0 || ^15.0.0",
     "prop-types": "^15.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -32,8 +32,7 @@
     "prop-types": "^15.0.0"
   },
   "devDependencies": {
-    "@types/react": "^0.14.55",
-    "babel": "^6.23.0",
+    "@types/react": "^15.0.21",
     "babel-cli": "^6.23.0",
     "babel-core": "^6.23.1",
     "babel-loader": "^6.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -49,12 +49,13 @@
...
-    "react": "^15.0.0",
-    "react-dom": "^15.0.0",
+    "karma-webpack": "^2.0.3",
+    "prop-types": "^15.5.4",
+    "react": "^15.5.3",
+    "react-dom": "^15.5.3",
     "rimraf": "^2.6.1",
     "safe-publish-latest": "^1.1.1",
-    "webpack": "^1.14.0"
+    "webpack": "^2.3.3"
```
jmeas: Is webpack still used by this project?

lencioni: I think it is used along with karma for testing.

package.json
```diff
@@ -28,7 +28,7 @@
   "author": "Brigade Engineering",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0"
```
lencioni: I think this might be considered a breaking change, which would require a semver major bump.

jmeas: Is one of the changes in this PR making the code incompatible with 0.14? From what I understand the `prop-types` lib will work fine with 0.14. If that's the case, then maybe we should keep the `peerDeps` open until the need arises to make it more strict?

trotzig: Yeah maybe this was too aggressive. What do you think @lencioni? You're right @jmeas that there's nothing right now that will break. One issue though is that we're not really testing with react 14. So if things go south, we'll likely find out about it from an issue. And at that time, it's going to be a little hard to say "oh, you know what, we'll just cut support for React 14". :man_shrugging:

lencioni: I think it is unlikely that we will break React 14 before React 16 is out, and even then we are probably pretty safe. I guess my preference would be to continue to support it at least until 16 is out or we have a compelling reason to remove support.

trotzig: Cool, I'll revert that commit before merging.

Issue #177, opened 2017-04-11
Call out in Twitter Lite post


lencioni:
> Up until recently, we were using react-waypoint, which worked well for us. However, in chasing the best possible performance for one of the main underlying components of our application, it just wasn’t fast enough.

> Waypoints work by calculating many different heights, widths, and positions of elements in order to determine your current scroll position, how far from each end you are, and which direction you’re going. All of this information is useful, but since it’s done on every scroll event it comes at a cost: making those calculations causes jank–and lots of it.

https://medium.com/@paularmstrong/twitter-lite-and-high-performance-react-progressive-web-apps-at-scale-d28a00e780a3

Is it possible that we've fixed these problems or do they still exist and if so are they fixable?

lencioni:
It looks like some of the changes they've made have been outlined here: http://itsze.ro/blog/2017/04/09/infinite-list-and-react.html

paularmstrong:
Hope this didn't spring up any incorrect assumptions. I wanted to point out that react-waypoints were great for us. However, the problem was that with them, for infinite scrolling lists as big and complex as ours, we had to do too many calculations and queries that caused re-render/re-paints in the browser.

By no means does that mean anyone that is using this package should be concerned. The findings we had were localized to *our application* and won't necessarily be reflected in another.

lencioni:
Thanks for the context @paularmstrong. I just want to make sure that if there are any performance opportunities for this project that we are taking them--whether they are in code improvements, better defaults, or documentation.

Looking at your screenshot and waypoint code, this particular bit of jank seems to happen because of the getBounds call in handleScroll. https://github.com/brigade/react-waypoint/blob/b23770d57fa66b4e20eeb3e8d80fa60d85135f36/src/waypoint.jsx#L161

It would be really helpful for me to know what version of react-waypoint you were using when you did your profiling.

lencioni:
Also thanks for the article. We've been doing a bunch of performance work at Airbnb and have done a lot of similar things that you've outlined. Got a long way to go, but we are making progress all the time!

paularmstrong:
> It would be really helpful for me to know what version of react-waypoint you were using when you did your profiling.

We had a fork from around v2. I don't really remember the context for why we forked it, as that was a loooong time ago. However, that `getBounds` really was the biggest hindrance to us. We already have to calculate the height of every single item rendered into a timeline up front, so doing an extra call that triggers jank is problematic and unnecessary.

[...]

PR #178, opened 2017-04-12
Add performance test page

trotzig:
I want to understand if there are any performance tweaks we can do to
make react-waypoint use fewer browser resources. Before I do that, it
helps having a page rendered with many waypoints, to dive into things
like fps, repaints, etc using the Chrome dev tools.

See #177 for more context.


.importjs.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  importDevDependencies: true,
+  environments: ['node', 'browser'],
+  declarationKeyword: 'import',
+};
```
jmeas: I've never seen this type of file before – is this a webpack feature?

lencioni: It's an editor plugin that @trotzig and I hack on. Check it out! https://github.com/galooshi/import-js

jmeas: Ah, cool! I will check it out.

Do you think it could be worthwhile adding a file comment linking to that project? I wasn't able to pull up that project from my google searches (although it very well could be that i'm bad at googling). No worries either way.

lencioni: Yep, I think this is a good idea.

webpack.config.js
```diff
@@ -0,0 +1,23 @@
+const path = require('path');
+
+const config = {
+  entry: path.join(__dirname, 'spec/performance-test.jsx'),
```
lencioni: Is it maybe confusing that this webpack config is only used for the performance test? I wonder if we should adjust the name of this file?

jmeas: I had a similar thought. I'm 👍 to renaming the file.

trotzig: This crossed my mind as well. I'll make the change.

jmeas:
This looks good to merge to me.

Issue #108, opened 2016-07-10
topOffset & bottomOffset


namwkim:
Thanks for the useful library! 

I wonder if topOffset & bottomOffset are really working? I am using "window" as an ancestor container and I set "80%" to topOffset. Does it mean that, when my scroll is at the bottom 20% of the "window" container, the waypoint will enter? It currently enters when the scroll reaches the bottom Please update the documentation if possible. 
Thanks!
Nam

jmeas:
I'd expect it to work like how you're describing. It should be straightforward to debug where things are diverging from your expectation with a breakpoint in this lib's source code. Do you have the time to do that?

jmeas:
@namwkim , I think things are working as I'd expect here.

The `topOffset` only affects the top boundary of the page – not the bottom. So the behavior you're seeing where elements are entering or leaving when they reach the bottom of the page seems like what I'd expect. Setting `topOffset: 20%` means that elements are considered "inside" the viewport when they are between 80% and 100% of the page size.

If this doesn't make sense, then allow me to try to summarize the algorithm behind this lib:

There's the concept of a **top boundary** and a **bottom boundary** in React Waypoint. When you're between the boundary points, you are considered **inside**. When you are not within the boundaries, you are **outside**. The default top and bottom boundary is 0, which will represent the edge of the viewport when the ancestor is the window.

The **enter** callback is fired when an element transitions to being **inside**. The **leave** event is triggered when an element transitions to being **outside**. These events fire at both boundaries: the top and the bottom boundary.

The `topOffset` and `bottomOffset` can be used to offset the top and bottom boundaries, respectively. By moving the `topOffset`, you've only changed the behavior of that top boundary. This means that you will still see enter and leave events when elements scroll to the bottom of the viewport (remember: the bottom boundary is controlled by the `bottomOffset`). However, setting `topOffset` means that you'll also see enter and leave events when elements pass the 80% viewport mark.

**A demo**

Here's a short gif showing the behavior when `scrollableAncestor` is `window`, and `topOffset` is `"80%"`. The grey square represents the bottom 20% of the viewport.

![waypoint](https://cloud.githubusercontent.com/assets/2322305/16938759/0ea0c634-4d30-11e6-814a-1ca670384657.gif)

(you can ignore the weird rendering glitch in the gif that causes the bg of the page to turn light grey sometimes. Not sure what that is 😛 )

You can play with this example over [here](https://github.com/jmeas/react-waypoint-issue-8-example). To install:

[...]

PR #183, opened 2017-04-29
Add offset and boundary images to the README

jmeas:
_Forever ago_ in #108, I drew up some images to show how the offsets and boundaries work. @trotzig recommended I add them to the README, and well, I intended to but never did. Whoops.

This finally adds them. There are probably improvements that could be made, but maybe this is a good place to start. What do y'all think?

//cc @xeophin


trotzig:
This is a good start, it will likely help a few people out. 👍

PR #184, opened 2017-04-29
Restrict lower bound of React to v0.14.9

### LINKED ISSUE #179: v6.0.0+: Restrict React 0.14 versions?
I haven't tested this, but this tweet, https://twitter.com/dan_abramov/status/852193956322521089 had me thinking that we may need to restrict the React 0.14 support post-v6, given that we've begun using `prop-types` in that version.

jmeas:
Depending on how broken the `prop-types` lib is in `< 0.14.9`, I could see this going out as either a major release, or a bug fix. Because they had to cut a new React version, my guess is that it's probably pretty broken, so I could see it going out as a bug fix.

Has anyone experienced using `prop-types` in `< 0.14.9`?

---

Resolves #179

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "author": "Brigade Engineering",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
+    "react": "^0.14.9 || ^15.0.0",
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks!

PR #185, opened 2017-05-01
Move prop-types to a proper dependency

trotzig:
This resolves some of the discussion in #179. The prop-types package is
small enough to not worry about duplicates ending up in peoples
(webpack/browserify) bundles. Also, other projects are also taking the
same ease-of-installation+dupes-don't-matter approach:

  - React Router
  https://github.com/ReactTraining/react-router/blob/ed6f3de4ea8baf950e65fd35e6785e0b493fb9e6/packages/react-router/package.json#L44

  - Redux-form
  https://github.com/erikras/redux-form/blob/4e7f5cfe5f34d2cf02cca841366d58196d330e35/package.json#L64

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,6 @@
   ],
   "dependencies": {
     "consolidated-events": "^1.0.1",
-    "prop-types": "^15.5.8"
+    "prop-types": "^15.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -25,13 +25,11 @@
     "testonly": "karma start --single-run",
     "testonly:watch": "karma start --no-single-run",
     "performance-test:watch": "webpack --watch --config webpack.config.performance-test.js"
-
   },
   "author": "Brigade Engineering",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^0.14.9 || ^15.0.0",
-    "prop-types": "^15.0.0"
+    "react": "^0.14.9 || ^15.0.0"
   },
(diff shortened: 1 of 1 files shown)


trotzig:
I think we should publish this (along with #184 that I just merged) as a minor fix. Does that sound like the right move?

jmeas:
What's your reasoning behind the minor release, @trotzig ? I was thinking either a bug fix or a major release, with a slight preference for a major release (despite the fact that we _just_ major'd)

package.json
```diff
@@ -67,6 +64,7 @@
     "onscroll"
   ],
   "dependencies": {
-    "consolidated-events": "^1.0.1"
+    "consolidated-events": "^1.0.1",
+    "prop-types": "^15.5.8"
```
trotzig: I think it was a mistake to specify such a new version here. I think I'll change to `^15.0.0` instead, so that we won't cause duplicate packages for people on older versions.

trotzig:
I don't know... I guess I saw the change in dependencies as a new "feature", but added in a way that was backwards-compatible (it's okay if people leave the prop-types package as a dependency of their own). But I guess your fix to the `react` peer dependency is a sort of breaking change (to a small amount of people, but still). 

I'll make it a major version just to be safe. Anything else we want to include in that release? Now's the chance! :)

package.json
jmeas: 👍

Issue #182, opened 2017-04-27
Add scrollableAncestor to debug output.


marlonpp:
I came across this weird behavior where if when the entry point of my application is the page where I have Waypoints it works fine, but, if the entry point is a different page that leads to a page where the Waypoints are, it doesn't.
What fixed the issue was ```scrollableAncestor={window}```, but it took me some time to figure out that because there is no log output for scrollableAncestor.

Would be great to have this info as part of the debug output.

trotzig:
I think this is a good idea. I've seen this come up in the past as well (https://github.com/brigade/react-waypoint/issues/171 comes to mind). Would you be willing to make that change?

marlonpp:
My bad, this already exists.
https://github.com/brigade/react-waypoint/blob/master/src/waypoint.jsx#L49

Only thing that could change is, removing `process.env.NODE_ENV !== 'production'`
from [L48](https://github.com/brigade/react-waypoint/blob/master/src/waypoint.jsx#L48).

jmeas:
👋  @marlonpp !

> Only thing that could change is, removing process.env.NODE_ENV !== 'production'
> from L48.

You'd want these logs in production?

marlonpp:
That's not the case, this check is done [here](https://github.com/brigade/react-waypoint/blob/master/src/debugLog.js).

jmeas:
Ah okay, makes sense. Perhaps the intention behind the duplicate check is so that in prod it wouldn't even make the call to `debugLog` (since that LoC would be stripped by Uglify), but I'm not sure. Perhaps @lencioni or @trotzig can shed more light there.

lencioni:
Yep, that's the idea. https://github.com/brigade/react-waypoint/commit/d1d3ff9551d8b409e0b2a0a7588ca0f4098fc5f4

Issue #171, opened 2017-04-05
Waypoint doesn't fire after switching url


SnowDiamond:
I'm trying to implement infinite scroll with fetching data from server. On initial page load **with fully browser reload** everything works as expected, so waypoint fires every time when I scroll to page bottom. But if I go to page with infinite scroll from any another page (using react-router) then waypoint fires only once at page load, and never again when scrolling to page bottom. Could you please help me to figure out what can be the problem?

My code is:

```
import React from 'react'
import UserItem from './UserItem'
import Waypoint from 'react-waypoint'

export default (props) => {
  return (
    <div>
      <div className='users-search-result clearfix'>
        {props.users.map(user => <UserItem key={user.id} user={user} />)}
      </div>
      <Waypoint onEnter={props.onPageScroll}>
        <div className='loader' />
      </Waypoint>
    </div>
  )
}
```

trotzig:
Is this component remounted in between page transitions? If it's not, then the waypoint will not trigger on page load. You could potentially work around that with a `key` prop set on the waypoint, something unique for the page: 

```jsx
  <Waypoint key={this.props.foo} />
```

You can also avoid rendering the waypoint during load, which would cause it to unmount and remount again: 

```jsx
{!this.state.loading && <Waypoint onEnter={() => this.setState({ loading: true })} />
```

Let me know if this helps!

SnowDiamond:
@trotzig  thank you for your help! Component is remounted between transitions, so componentDidMount fires every time during transitions. 
I also tried to specify `key`, like you suggested but it doesn't help. Any other thoughts? :)

trotzig:
You could add the `debug` prop and look at what's going on in the console. It might be that it's always in view. 

```jsx
<Waypoint debug />
```

SnowDiamond:
Maybe I explain my problem not clear. Problem is that waypoint **doesn't fires at all** when I scroll to page bottom, so nothing to debug. And this happens only if I go to the page from any another page.  But if I fully reload browser on the page it works correct, and fires each time when I scroll to the bottom.

trotzig:
I re-read your initial comment now. That was clear from the start, I was just a little quick to come to conclusions. Sorry!

The `debug` thing  is still interesting though, it should give us a little more input. Try with a full page reload and with a react-router transition, then compare the different logs.

SnowDiamond:
Debug log with full browser reload:
```
["scrollableAncestor", Window]
[...]

PR #187, opened 2017-05-03
Assume `window` as scrollableAncestor for <body>

trotzig:
We've seen reports (e.g. [1], [2]) where people have to add
`scrollableAncestor={window}`, because their <body> element has an
overflow style that is either "auto" or "scroll". Unless I'm mistaken,
there's no way to make the body scroll independently, which means that
we can assume that the element to attach scroll listeners to is
`window`.

Making this assumption led to a bit of simplification around a few
exceptions we had made when traversing up the DOM tree. Once we reach
`document.body`, we simply return. I can't think of a scenario where we
would want to continue traversing from that node.

[1] #171
[2] #182

### DIFF
--- a/spec/waypoint_spec.js
+++ b/spec/waypoint_spec.js
@@ -1138,9 +1138,17 @@ describe('<Waypoint>', function() {
       this.parentStyle.height = 'auto';
       this.parentStyle.overflow = 'visible';
 
+      // This is only here to try and confuse the _findScrollableAncestor code.
+      document.body.style.overflow = 'auto';
+
       // Make the spacers large enough to make the Waypoint render off-screen
       this.topSpacerHeight = window.innerHeight + 1000;
       this.bottomSpacerHeight = 1000;
+    });
+
+    afterEach(() => {
+      // Reset body style
+      document.body.style.overflow = '';
     });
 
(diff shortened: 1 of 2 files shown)


trotzig:
I wanted to get this in before carving a new major release.

spec/waypoint_spec.js
```diff
@@ -1138,11 +1138,19 @@ describe('<Waypoint>', function() {
...
+      document.body.style.overflow = 'auto';
+
       // Make the spacers large enough to make the Waypoint render off-screen
       this.topSpacerHeight = window.innerHeight + 1000;
       this.bottomSpacerHeight = 1000;
     });
 
+    afterEach(() => {
+      // Reset body style
+      document.body.style.overflow = '';
```
jmeas: This should probably reset it to whatever it was before Waypoint got involved, rather than an empty string

trotzig: Ah! This is the "right" way to do things. I initially tried `delete` and setting to `undefined` here before. 

http://stackoverflow.com/questions/3506050/how-to-reset-the-style-properties-to-their-css-defaults-in-javascript

jmeas: Oh, cool – I learned something new. Never mind, then!

PR #188, opened 2017-05-05
Improve startup time

lencioni:
While doing some profiling, I noticed that on pages that render a lot of
waypoints, a lot of time is spent in componentDidMount. This is mostly
due to the cost of calling setTimeout, but also because of the other
work we are doing in here, like findScrollableAncestor.

By consolidating the setTimout calls, we greatly reduce the cost of
componentDidMount. We can take this a step further and defer
initializing the event listeners and finding the scrollable ancestors
since we won't need any of that anyway until our initial handleScroll
call can happen. This effectively makes componentDidMount free, which
frees up the main thread for more important work, like rendering pixels.

I believe that this also allows the calls to findScrollableAncestor to
work faster by avoiding some layout thrashing. Since this function
traverses up the DOM tree and computes styles on a bunch of nodes, this
is important. When these operations happened in componentDidMount, it is
possible that other components would invalidate layout in between these
calls from multiple waypoints. By clustering all of them together, we
are less likely to have to recalculate style or compute layout every
time.

Additionally, there may be cases where waypoints are rendered and
unmounted very quickly. In this case, we might actually be able to avoid
some of the unnecessary work we are doing here, which will be a small
win for those situations.

The main drawback to this is that if you have a lot of waypoints on a
page, the previous approach keeps each initial run a bit more async,
which allows the browser to do work in between. Now, all of that work is
in the same block. I think this is an acceptable tradeoff.

Here are some flame charts I generated on a page that has a lot of waypoints (close to 100 I think).

componentDidMount Before (most of those orange blocks are calls to setTimeout):

screen shot 2017-05-05 at 10 45 49 am

componentDidMount after consolidating setTimeout only:

screen shot 2017-05-05 at 10 45 41 am

componentDidMount after moving everything into the callback (almost all of the time left is spent in other components that are not waypoint):

screen shot 2017-05-05 at 10 46 00 am

timeouts being called before this change:

screen shot 2017-05-05 at 10 48 04 am

timeouts being called after this change:

screen shot 2017-05-05 at 10 48 31 am

### DIFF
--- a/src/onNextTick.js
+++ b/src/onNextTick.js
@@ -1,0 +1,36 @@
+let timeout;
+const timeoutQueue = [];
+
+export default function onNextTick(cb) {
+  timeoutQueue.push(cb);
+
+  if (!timeout) {
+    timeout = setTimeout(() => {
+      timeout = null;
+
+      // Drain the timeoutQueue
+      let item;
+      // eslint-disable-next-line no-cond-assign
+      while (item = timeoutQueue.shift()) {
+        item();
+      }
+    }, 0);
+  }
+
+  let isSubscribed = true;
+
+  return function unsubscribe() {
+    if (!isSubscribed) {
+      return;
+    }
+
+    isSubscribed = false;
+    const index = timeoutQueue.indexOf(cb);
+    timeoutQueue.splice(index, 1);
+
+    if (!timeoutQueue.length && timeout) {
+      clearTimeout(timeout);
+      timeout = null;
+    }
+  };
+}
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -7,6 +7,7 @@ import constants from './constants';
 import debugLog from './debugLog';
 import ensureChildrenIsSingleDOMElement from './ensureChildrenIsSingleDOMElement';
 import getCurrentPosition from './getCurrentPosition';
+import onNextTick from './onNextTick';
 import resolveScrollableAncestorProp from './resolveScrollableAncestorProp';
 
 const defaultProps = {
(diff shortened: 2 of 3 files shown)


lencioni:
Looks like this breaks tests. I'll wait until I get your feedback before I bother updating tests.

src/onNextTick.js
```diff
@@ -0,0 +1,36 @@
+let timeout;
+const timeoutQueue = [];
+
+export default function onNextTick(cb) {
```
lencioni: I wonder if this would be better named `defer`?

src/waypoint.jsx
```diff
@@ -42,34 +43,33 @@ export default class Waypoint extends React.Component {
...
+        this._handleScroll,
+        { passive: true }
+      );
 
-    // this._ref may occasionally not be set at this time. To help ensure that
-    // this works smoothly, we want to delay the initial execution until the
-    // next tick.
-    this.initialTimeout = setTimeout(() => {
-      this.initialTimeout = null;
       this._handleScroll(null);
```
lencioni: I wonder if this should be wrapped in a second `onNextTick`?

src/onNextTick.js
trotzig: Either or. `onNextTick` is slightly more revealing if you ask me.

src/waypoint.jsx
trotzig: Why? 😃

trotzig:
Thanks for being so thorough! In case you're waiting for my blessing before fixing up those tests, here it is: 🙏

src/onNextTick.js
lencioni: That's true, but I think it is not actually accurate because of the way the JS event loop functions.

src/waypoint.jsx
lencioni: I was thinking that, depending on what happens in `handleScroll`, there might be some layout thrashing that happens and we might be able to reduce that by moving that work out. Or, maybe we should put the calls to `onEnter`, `onLeave`, and `onPositionChange` in `handleScroll` in an `onNextTick` callback. But I worry that might cause some confusing behavior and should perhaps be left up to the consumer to avoid thrashing if it comes up.

lencioni:
@jmeas do you have any thoughts on this type of change?

jmeas:
Been busy lately. I'm :+1: after a quick skim of the code and this PR, but I haven't dug into it too much yet :)

PR #191, opened 2017-05-11
Add tests for onNextTick()

lencioni:
We recently had a bug in this function that we fixed in #190, so I wrote
some tests to prevent it from regressing. At the same time I decided to
add an optimization to avoid the array scan after the first time
unsubscribe is called.

cc @ssnau

### DIFF
--- a/spec/onNextTick_spec.js
+++ b/spec/onNextTick_spec.js
@@ -1,0 +1,135 @@
+import onNextTick from '../src/onNextTick';
+
+describe('onNextTick()', () => {
+  beforeEach(() => {
+    jasmine.clock().install();
+  });
+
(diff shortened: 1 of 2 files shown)


spec/onNextTick_spec.js
```diff
@@ -0,0 +1,135 @@
...
+    unsub();
+    unsub();
+    unsub();
+
+    jasmine.clock().tick(1);
+
+    expect(called).toEqual([0, 2]);
+  });
+
+  it('does nothing when unsubscribing a callback that has already been called', () => {
```
lencioni: I verified that this test reproduces the bug that we recently fixed.

PR #218, opened 2017-09-26
Remove Node 5 from Travis matrix

jmeas:
Following up from https://github.com/brigade/react-waypoint/pull/217

This might seem arbitrary, but let me explain. If you check out [the Node release schedule](https://github.com/nodejs/Release), you can see that Node 4/6/8 are either active or LTS. This is why I've kept those around.

5 and 7 are not, but 7 is a pretty recent version of Node, so I kept that around, too. I think we're safe to remove Node 5 testing. Honestly, we could probably even get away with no longer testing the 4 versions, too, but I opted for leaving it for now. It's out of maintenance support in mid-2018, so maybe that's a better time to stop testing it?


trotzig:
Good call.

PR #219, opened 2017-09-27
Upgrade peerDeps to react 16

jnsdls:
since react-waypoint actually works fine with react16


jnsdls:
ah, sorry just saw #215, which is solving the same issue just better :)

PR #220, opened 2017-10-02
Build with rollup

lencioni:
Previously, we were building waypoint just by running it through Babel.
Since we split things out into separate modules, this caused a bit of
extra overhead in the bundle size and at runtime due to the extra
wrapping of these modules.

Thankfully, rollup was built to help us produce a smaller build. To make
this work, I needed to remove the Babel configuration that specifies a
module transformer, so I decided to switch to the Airbnb Babel preset,
which exposes an option for this (and under the hood uses
babel-preset-env, which we need to migrate to anyway). More info:

  https://github.com/rollup/rollup-plugin-babel#configuring-babel

The rest of this was loosely based off the rollup-starter-lib
configuration. This gives us an ESM build pretty much for free.

  https://github.com/rollup/rollup-starter-lib/tree/babel

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "name": "react-waypoint",
   "version": "7.2.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
-  "main": "build/index.cjs.js",
-  "module": "build/index.esm.js",
+  "main": "build/index.js",
+  "module": "build/index.mjs",
   "types": "index.d.ts",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "react-waypoint",
   "version": "7.2.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
-  "main": "build/waypoint.js",
+  "main": "build/index.cjs.js",
+  "module": "build/index.esm.js",
   "types": "index.d.ts",
   "repository": {
     "type": "git",
@@ -11,7 +12,7 @@
   "homepage": "https://github.com/brigade/react-waypoint",
   "bugs": "https://github.com/brigade/react-waypoint/issues",
   "scripts": {
-    "build": "npm run clean && babel src/ -d build/",
+    "build": "npm run clean && rollup -c",
     "check-changelog": "expr $(git status --porcelain 2>/dev/null| grep \"^\\s*M.*CHANGELOG.md\" | wc -l) >/dev/null || (echo 'Please edit CHANGELOG.md' && exit 1)",
     "check-only-changelog-changed": "(expr $(git status --porcelain 2>/dev/null| grep -v \"CHANGELOG.md\" | wc -l) >/dev/null && echo 'Only CHANGELOG.md may have uncommitted changes' && exit 1) || exit 0",
     "clean": "rimraf build",
@@ -36,9 +37,7 @@
     "babel-cli": "^6.23.0",
     "babel-core": "^6.23.1",
     "babel-loader": "^6.4.0",
-    "babel-plugin-add-module-exports": "^0.2.1",
-    "babel-plugin-transform-react-jsx": "^6.23.0",
-    "babel-preset-es2015": "^6.22.0",
+    "babel-preset-airbnb": "^2.4.0",
     "eslint": "^3.17.1",
     "eslint-config-brigade": "^3.2.1",
     "eslint-plugin-react": "^6.10.0",
@@ -53,6 +52,8 @@
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "rimraf": "^2.6.1",
+    "rollup": "^0.50.0",
+    "rollup-plugin-babel": "^3.0.2",
     "safe-publish-latest": "^1.1.1",
     "webpack": "^2.3.3"
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,21 @@
+import babel from 'rollup-plugin-babel';
+import pkg from './package.json';
+
+export default [
+  {
+    entry: 'src/waypoint.jsx',
+    external: [
+      ...Object.keys(pkg.dependencies),
+      ...Object.keys(pkg.peerDependencies),
+    ],
+    targets: [
+      { dest: pkg.main, format: 'cjs' },
+      { dest: pkg.module, format: 'es' }
+    ],
+    plugins: [
+      babel({
+        exclude: ['node_modules/**']
+      })
+    ]
+  }
+];
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,14 +3,14 @@ import pkg from './package.json';
 
 export default [
   {
-    entry: 'src/waypoint.jsx',
+    input: 'src/waypoint.jsx',
     external: [
       ...Object.keys(pkg.dependencies),
       ...Object.keys(pkg.peerDependencies),
     ],
-    targets: [
-      { dest: pkg.main, format: 'cjs' },
-      { dest: pkg.module, format: 'es' }
+    output: [
+      { file: pkg.main, format: 'cjs' },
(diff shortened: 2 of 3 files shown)


lencioni:
I haven't used rollup before, so it is likely I missed something important. Please keep this in mind while reviewing.

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "react-waypoint",
   "version": "7.2.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
-  "main": "build/waypoint.js",
+  "main": "build/index.cjs.js",
+  "module": "build/index.esm.js",
```
ljharb: If you want to be compatible with what node will support, you'll want to do `build/index.js` and `build/index.mjs` here.

jmeas:
Good idea here, @lencioni !

I'll give a more thorough review tonight, but I thought I'd pop in here to suggest that we stick with `.js` for now, even for the `module` field in package.json.`.mjs` support in Node is experimental, so I think we should wait until the spec is finalized, and the flag is removed, before we include `.mjs` files in this project. Just my two cents :v:

jmeas:
Overall, this looks good. Two questions:

Do we want to export in the UMD format for browser-users? We could do that by adding `format: 'umd',` to the Rollup config, as well as using the `globals` option, which is a map of globals variables to the library names that we import.

Also, for the ES modules build, what do we get out of running it through Rollup? Could we just run the lib through Babel? [This is the approach that Redux takes](https://github.com/reactjs/redux/blob/9e51c170697b9b86e85dffa076549546eff05ab0/package.json#L23). Would this avoid any `import * as` Rollup issues that @ljharb mentioned?

ljharb:
Regardless of the module field, `.mjs` in node is a foregone conclusion - if you want to omit that field, fine, but an `.mjs` file is the proper way moving forward for the JS ecosystem, full stop.

jmeas:
Sure, but it is still currently an experimental feature. It'll likely work out fine either way, but we should just be aware of the fact that we're adding in a thing that depends on a feature that is, at this time, experimental. If we're cool with that then :+1:

ljharb:
Fair point. In that case, there shouldn't be a "module" field at all because that's not a feature that's standardized in any way whatsoever :-)

jmeas:
> In that case, there shouldn't be a "module" field at all because that's not a feature that's standardized in any way whatsoever :-)

I agree with you that we may not need the `module` field, but not because it's not standardized. You may already know this, but common build tools [recognize this field](https://github.com/rollup/rollup/wiki/pkg.module), and will prefer it over `main`.

This can be useful for tree shaking, but given this library's single export, I don't believe that there is anything to shake. For this reason, we could probably just get away with just exporting a single UMD build as `main`.

ljharb:
I'm not sure why a UMD build would even be needed - just a single CJS file could work too.

jmeas:
> I'm not sure why a UMD build would even be needed - just a single CJS file could work too.

You're probably right. I don't have any numbers on this, but I suspect that the number of React developers who aren't using _some_ kind of build tool is miniscule. Maybe bundling UMD is a holdover from some previous time, but I still find myself doing it. It's a simple thing for me to add to my libraries, and it _could_ help a developer trying to consume the library.

I'd be fine with us just shipping a CJS file as `main`, though.

trotzig:
It looks like build failures are spurious. I've manually retriggered the failing ones.

lencioni:
> Also, for the ES modules build, what do we get out of running it through Rollup? Could we just run the lib through Babel?

I think we get two things. The first is that it cuts down on the amount of work that app bundlers will need to do at build time. The second is that it makes all of the rolled up and not exported functions private to the module, which gives us more confidence when deciding what is a breaking change or not.

>  In that case, there shouldn't be a "module" field at all

I'd like to keep the module field because at Airbnb, we are working on improving our builds in ways that will be able to take advantage of this. I don't care if the file is called index.mjs or spaghetti.and.meatballs. We can change it later if we need to without much hassle. I think index.mjs is fine. Are you comfortable with this @jmeas? If not I'm happy to change it back to index.esm.js.

> This can be useful for tree shaking, but given this library's single export, I don't believe that there is anything to shake. For this reason, we could probably just get away with just exporting a single UMD build as main.

I think it can also be useful for scope hoisting. It's true that we only have one export, but we do import a few other things that may benefit from some of these optimizations. And I believe that those imports need to be using `import` and not `require` (which is the only real difference between the main and module versions that rollup produces here).

If any tree shaking could happen in this package, I think rollup would be shaking it out here.

> I'm not sure why a UMD build would even be needed - just a single CJS file could work too.

Agreed, I don't think we need UMD. It's not something we have now and nobody has asked for it. We can always add it later, but that is tangential to what I am hoping to accomplish here.

jmeas:
Ah, I should have checked – I just assumed we were currently already bundling a UMD build. I agree that it’s outside of the scope of this PR to add that in.

I also agree with the other points you mentioned. As long as we’re absolutely certain that we don’t want to go with `spaghetti.and.meatballs`, let’s get this thing merged!

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "react-waypoint",
   "version": "7.2.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
-  "main": "build/waypoint.js",
+  "main": "build/index.js",
```
ljharb: if both of these are just "build" - which is the best practice - does that mean rollup can't figure out the destination filenames?

lencioni: I don't understand your question. What is a best practice? What is rollup not figuring out?

ljharb: `"main": "build", "module": "build"`.

I don't see anywhere else `index.js` or `index.mjs` are specified tho, which leads me to believe that rollup infers the filenames from the package.json fields.

lencioni: No, it doesn't infer anything. It is configured with an explicit dependency on these fields in the rollup config. https://github.com/brigade/react-waypoint/pull/220/files#diff-ff6e5f22a9c7e66987b19c0199636480

```diff
+    targets: [
+      { dest: pkg.main, format: 'cjs' },
+      { dest: pkg.module, format: 'es' }
+    ],
```

ljharb: That specifies the format, but not the file extension. Where would spaghetti and meatballs go?

lencioni: It specifies the path currently. If we change it to spaghetti and meatballs, it would go in the package.json.

rollup.config.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default [
+  {
+    entry: 'src/waypoint.jsx',
+    external: [
+      ...Object.keys(pkg.dependencies),
+      ...Object.keys(pkg.peerDependencies),
+    ],
+    targets: [
+      { dest: pkg.main, format: 'cjs' },
```
jmeas: I just looked this up after reading @ljharb 's questions, and I don't see `targets` in [the Rollup docs](https://rollupjs.org/#big-list-of-options). Should this instead be `output` and `file`, rather than `targets` and `dest`?

jmeas: Looks the answer might be [yes](https://github.com/rollup/rollup/blob/6fc2088bfaf6b47b7225223398d1335f167a0aea/src/utils/deprecateOptions.js#L33-L34)

package.json
ljharb: Right - that confirms my initial hunch, which is that putting just `"build"` wouldn't work, because it would deprive rollup of the actual path.

rollup.config.js
ljharb: Ooh, in that case yes that's better :-D

lencioni: Ah, good catch. I added a commit to convert these.

lencioni: I looked into updating the code I pulled this from and it looks like there is an open PR for it already! https://github.com/rollup/rollup-starter-lib/pull/2

jmeas:
🎉

PR #223, opened 2017-10-05
Make ES module opt-in

### LINKED ISSUE #221: Error after updated to v7.3.0
Hi,

I have updated react-waypoint to 7.3.0 today and I am getting this error:

screen shot 2017-10-04 at 11 51 59 am

Everything is working fine after I revert back to v7.2.0

Is this a bug or am I missing something?

Thanks!

trotzig:
After releasing 7.3.0, we've seen folks run into issues (#221) with the
new `module` field added to package.json. I've tried to read up on how
webpack uses this field, and it looks like it will pick it up by
default. This is tricky because most people will expect a CommonJS
export (i.e. `module.exports = ...`), but now they suddenly have to
import react-waypoints using

  const Waypoint = require('react-waypoint').default;

There's a lengthy issue over at webpack discussing this [1], with a few
proposed solutions at the bottom:

- offer a separate package like lodash does (lodash-es), or
- skip the module field and use a es-specific entry point, like import a from 'yourpkg/es'
- skip the module field, avoid es modules (my preferred solution for most of my modules)

I'm opting for the second point here. We can use the commonjs bundle as
the default, and then if people are adventurous they can point at the es
module directly:

  import Waypoint from 'react-waypoint/build/index.mjs';

We're not really going to the bottom of this issue with this commit. But
I believe it will solve headaches for a lot of applications. For
instance, I just found out that one developer in my team spent time
chasing down a bug where parts of the application wouldn't work.
Waypoints that were imported with `import Waypoint from
'react-waypoint'` were working, but not the ones being imported via
`const Waypoint = require('react-waypoint')` (we have some legacy).

Fixes #221

[1] https://github.com/webpack/webpack/issues/4742

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-waypoint",
   "version": "7.3.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
-  "main": "build/index.js",
+  "main": "cjs/index.js",
   "types": "index.d.ts",
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,7 +3,6 @@
   "version": "7.3.0",
   "description": "A React component to execute a function whenever you scroll to an element.",
   "main": "build/index.js",
-  "module": "build/index.mjs",
```
lencioni: I believe the rollup config has a direct dependency on this field, so you will need to modify that as well.

trotzig: Ah, thanks. I had missed that.

trotzig:
I agree @jmeas, let me make that change before I merge.

trotzig:
I'm going to push this as a patch version. It could be argued that it should be major (since it's a breaking change) but I'll rather break things for people who had just adopted the es build than to force everyone on ^7.0.0 (and are now in a semi-broken state) to update to 8.

PR #230, opened 2017-12-04
Update typescript definition

rosskevin:
TS definition was using old style declaration, which was also not exporting the namespace.  Use this as `import * as Waypoint from 'react-waypoint'` and it allows access to everything in the namespace e.g. `CallbackArgs`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,15 @@
-/// <reference types="react" />
+import * as React from 'react';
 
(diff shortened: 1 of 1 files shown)


lencioni:
@fongandrew @nicolas-schmitt could you please review this PR when you have a moment?

fongandrew:
LGTM.

trotzig:
Thanks @rosskevin!

PR #231, opened 2017-12-14
Add second arg to Typescript component definition

eessex:
JSX elements in Typescript require two args when defined, props and state.  This provides a null arg for state, killing the below Typescript errors:

_Happens when the library is required-_
`[at-loader] ./node_modules/react-waypoint/index.d.ts:3:32
    TS2314: Generic type 'Component<P, S>' requires 2 type argument(s).`

_Happens when library is used -_ 
`[at-loader] ./src/Components/Publishing/Nav/Nav.tsx:73:10
    TS2604: JSX element type 'Waypoint' does not have any construct or call signatures.
`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-declare class Waypoint extends React.Component<Waypoint.WaypointProps> {
+declare class Waypoint extends React.Component<Waypoint.WaypointProps, null> {
   above: string;
   below: string;
   inside: string;


trotzig:
Thanks! I don't use typescript myself, so I'm making the assumption you know what you are doing. 😃

zamiang:
Seems to have created a new problem for me 

```
infinite-scroll.tsx(46,9): error TS2605: JSX element type 'Waypoint' is not a constructor function for JSX elements.
  Types of property 'setState' are incompatible.
    Type '{ <K extends never>(f: (prevState: null, props: WaypointProps) => Pick<null, K>, callback?: (() =...' is not assignable to type '{ <K extends never>(f: (prevState: Readonly<{}>, props: any) => Pick<{}, K>, callback?: (() => an...'.
      Types of parameters 'f' and 'f' are incompatible.
        Types of parameters 'prevState' and 'prevState' are incompatible.
          Type 'null' is not assignable to type 'Readonly<{}>'.
         <Waypoint onEnter={this.getNextPage} />
```

eessex:
Hm maybe an empty object could do it

eessex:
@zamiang does it work for you if replacing null with {} ?

zamiang:
Yup - just submitted a pr https://github.com/brigade/react-waypoint/pull/232

PR #232, opened 2017-12-16
Fix typescript types from #231

zamiang:


trotzig:
Thanks!

trotzig:
Released in v7.3.4!

PR #234, opened 2017-12-22
Added module entry point in package.json

Andarist:
Webpack2+ and rollup will understand this out of the box when importing from the lib, no need to import from `'react-waypoint/es'`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "7.3.4",
   "description": "A React component to execute a function whenever you scroll to an element.",
   "main": "cjs/index.js",
+  "module": "es/index.js",
   "types": "index.d.ts",
   "repository": {
     "type": "git",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -10,7 +10,7 @@ export default [
     ],
     output: [
       { file: pkg.main, format: 'cjs' },
-      { file: 'es/index.mjs', format: 'es' }
+      { file: pkg.module, format: 'es' }
     ],
     plugins: [
       babel({


trotzig:
Thanks @Andarist! We used to have `module` in the package.json file, but it was causing issues for people. This commit removed it: https://github.com/brigade/react-waypoint/commit/0e791c0a71b01918319fcf7875552a3ceb1e0974. Here's the issue that was filed after we added it: https://github.com/brigade/react-waypoint/issues/221

I'm not going to close this PR just yet. If you or someone else can come up with a few workarounds for people who will run into trouble from this, I'm okay merging this and releasing it as v8 (so that people will have to make a manual upgrade + migration).

Andarist:
The problem described in the other issue was caused by using `.mjs` extension. Merging this PR as is wouldn't case any trouble.

Using `.mjs` as extension for `module`'s entry wouldn't give any benefits, `.mjs` is node's thing and `module` is targeting web bundlers.

jmeas:
I am a huge :+1: to this change as long as our build configuration outputs the right files.

This is what I do on personal libraries as well as libraries at work, and it works out amazingly. Furthermore, numerous other widely-used JS libraries do this, too, such as Redux. This configuration is the _de facto_ way to support tree shaking with webpack without any side effects caused by unsupported extensions such as `.mjs`.

I discouraged us from using `.mjs` in the original PR where we introduced `module`, and, perhaps unsurprisingly, that decision caused problems for our users. The problem has never been the `module` field (which is widely supported); it’s the extension (which has very little/nonexistent ecosystem support).

tl;dr, I am strongly in favor of this change, and holding off on `.mjs` until it is more widely supported :+1:

lencioni:
I agree with James. Let's ship module field and go to .js instead of .mjs

Although by now I think create react app supports mjs

trotzig:
Thanks for all the input folks!

PR #235, opened 2017-12-23
Let prop types be removable in consumers production bundles

Andarist:


jmeas:
Thanks for the PR! The description makes this seem like a good change, but I’ll need to do more research before I can give this a thumbs up or down.

I appreciate these PRs, @andarist!

Andarist:
Sure thing. If you have any questions, just ask :) To see the actual impact on the codebase I'd suggest building `master` and my branch and seeing the diff between them.

lencioni:
Although I love the intention here, I worry a bit about this breaking
folks. We should make sure that the proptypes are a named export to people
have a path forward if they need the proptypes for some reason.

Andarist:
Hm, yeah that could be done probably, although maybe better to export them if requested by the community and not before?

lencioni:
Yeah that might be alright too

trotzig:
I'm 👍 this change as well.

PR #237, opened 2018-01-09
Type Waypoint class properties as static

cecchi:
These properties are set [directly on the constructor](https://github.com/brigade/react-waypoint/blob/master/src/waypoint.jsx#L324-L327), and thus should be static. The current typings cause an error if you attempt to access those properties:

```
Property 'inside' does not exist on type 'typeof Waypoint'.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,10 @@
 import * as React from 'react';
 
 declare class Waypoint extends React.Component<Waypoint.WaypointProps, {}> {
-  above: string;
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks! Code looks good to me, just waiting to hear back on the travis build.

trotzig:
Not sure why Travis is taking so long. It looks green across the board though, so I'm going to go ahead and merge.

Issue #238, opened 2018-01-13
.default is back in 8.0


realityking:
Unfortunately adding the `modules` property to package.json in #234 brought #221 back. We're using webpack 3.10.0

This is especially painful when trying to use both Node.js for server side rendering and the browser. 

I fixed this by adding `mainFields: ['browser', 'main']` to our webpack config though that's obviously not optimal.

trotzig:
Ouch, sorry about that. I see why this would be painful in a project mixing commonjs requires and es imports. I've tried to find a solution to this but I'm coming up short. Do you know of any libraries out there that publish both a `main` and a `module` and still works with es and commonjs?

realityking:
The thing is we're not actually mixing them - our codebase is 100% CommonJS. The problem is that webpack doesn't play nice. 😞 

Given the current state of the ecosystem I think the only safe way to ship both ESM and CJS in the same package is to not have any default exports. If every export is a named export things work as expected.

MatthewHerbst:
Slightly related to #246

PR #239, opened 2018-01-16
Fix import error in typescript

jyhwng:
Fix `export` to es6 syntax in type definition. Used `default export`. If #238 is on purpose, consider this PR

([Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export))

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@ declare class Waypoint extends React.Component<Waypoint.WaypointProps, {}> {
   static invisible: string;
 }
 
-export = Waypoint;
+export default Waypoint;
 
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks for the fix! I don't have a fix for the issue introduced in #238 so at least for now this will make things better. 

I'm not using typescript myself, and I'm slightly embarrassed about the number of typescript related issues we introduce. Do you think you could help out by writing some kind of test to verify the typescript declarations? It could help prevent these issues going forward.

jyhwng:
@trotzig Thanks for merging :) Sure I'd be happy to contribute.

PR #254, opened 2018-05-25
New prop added - fullView

onyoon7:
if the waypoint is displayed on the screen fully, not partly.
The onEneter/onLeave fire

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,12 @@ enters or leaves the viewport. For details, see [Children](#children), below.
     fireOnRapidScroll: PropTypes.bool,
 
     /**
+     * fullView - if the waypoint is displayed on the screen fully, not partly.
+     * The onEneter/onLeave fire
+     */
+    fullView: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


MatthewHerbst:
Hi @onyoon7, thanks for the PR, and sorry for the delay in responding to it. @trotzig any objections to merging this?

onyoon7:
Thank you for reading about this Pull request and reviewing it in detail @jamesplease . 🙇‍♂️

Here are some reasons I suggested fullView:

- Run a callback if the child is completely visible
For example, when a customer saw a product on the eCommerce site, I realized that the customer was aware of the product and wanted to run a callback. In this case, onEnter and onLeave do not know if the entire child has been viewed.

- onEnter: executed when child starts to be visible
- fullView: run when the entire child is visible
- onLeave: run when child starts to be hidden

If you feel that the above features are unnecessary or have any good improvements, please feel free to do so.

Thank you for reading.

PR #257, opened 2018-06-18
Add node 10 to test matrix, remove node 4.2

lencioni:


jamesplease:
This is good. I'm thinking that it would probably be OK to just test on LTS versions (8 and 10) if we wanted to remove even more versions.

jamesplease:
This is good. I'm thinking that it would probably be OK to just test on LTS versions (8 and 10) if we wanted to remove even more versions.

lencioni:
One of the node 7 builds failed on the PR run, but it looks like it just got weirdly stuck and timed out. Same build passed on the push run so I'm going to merge this in.

Issue #263, opened 2018-07-06
componentDidUpdate forces synchronous layout, which causes layout thrashing


lencioni:
I was doing some profiling and I noticed that when Waypoint updates, we call `this._handleScroll` synchronously, which eventually calls functions that force layout.

<img width="767" alt="dev vacation rentals homes experiences places - airbnb 2018-07-06 08-16-08" src="https://user-images.githubusercontent.com/195534/42386689-0b6d5e14-80f5-11e8-88a5-231fe834348e.png">

https://github.com/brigade/react-waypoint/blob/b06941c20643afb8a3d427f7c8d79b6ba137f2bb/src/waypoint.jsx#L80-L92

Since this happens during render, the layout is immediately invalidated and then triggered again once rendering is completed.

We avoid this problem in `componentDidMount` by deferring this work until the next tick.

https://github.com/brigade/react-waypoint/blob/b06941c20643afb8a3d427f7c8d79b6ba137f2bb/src/waypoint.jsx#L47-L73

I think we should probably consider making two changes to Waypoint to help address this:

1. Extend `React.PureComponent` instead of `React.Component`. In the place I saw this problem, we were rendering a waypoint with pure props, so the update in this case could be completely avoided by using PureComponent.

```js
<Waypoint
  onEnter={this.onEnterWaypoint}
  scrollableAncestor={this.props.scrollableAncestor}
/>
```

2. Defer `this._handleScroll` in `componentDidUpdate` using `onNextTick`, like we do in `componentDidMount`. This will still require layout, but it will happen around the same time that the browser was going to do layout anyway, avoiding the thrash.

PR #264, opened 2018-07-06
Extend React.PureComponent instead of React.Component

lencioni:
I was doing some profiling and I noticed that when Waypoint updates, we
call `this._handleScroll` synchronously, which eventually calls functions
that force layout. Since this happens during render, the layout is
immediately invalidated and then triggered again once rendering is
completed.

We avoid this problem in `componentDidMount` by deferring this work
until the next tick.

I think we should consider making two changes to Waypoint to help
address this:

1. Extend `React.PureComponent` instead of `React.Component`. In the
   place I saw this problem, we were rendering a waypoint with pure
   props, so the update in this case could be completely avoided by
   using PureComponent.

   ```jsx
   <Waypoint
     onEnter={this.onEnterWaypoint}
     scrollableAncestor={this.props.scrollableAncestor}
   />
   ```

2. Defer `this._handleScroll` in `componentDidUpdate` using
   `onNextTick`, like we do in `componentDidMount`. This will still
   require layout, but it will happen around the same time that the
   browser was going to do layout anyway, avoiding the thrash.

This commit implements the first of these two interventions.

Partly addresses #263

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -22,10 +22,13 @@ const defaultProps = {
   fireOnRapidScroll: true,
 };
 
-/**
- * Calls a function when you scroll to the element.
- */
-export default class Waypoint extends React.Component {
+// React.PureComponent was added in React 15.3.0
+const BaseClass = typeof React.PureComponent !== 'undefined' ?
+  React.PureComponent :
+  React.Component;
+
+// Calls a function when you scroll to the element.
+export default class Waypoint extends BaseClass {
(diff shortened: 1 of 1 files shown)


jamesplease:
image

(Emphasized part is highlighted)

Would this be a breaking change, as all children components of a Waypoint must now also be pure?

lencioni:
@jamesplease @trotzig Any more thoughts on this or do you think we are good to merge?

PR #265, opened 2018-07-06
Defer `handleScroll` in `componentDidUpdate`

### LINKED ISSUE #263: componentDidUpdate forces synchronous layout, which causes layout thrashing

lencioni:
I was doing some profiling and I noticed that when Waypoint updates, we
call `this._handleScroll` synchronously, which eventually calls functions
that force layout. Since this happens during render, the layout is
immediately invalidated and then triggered again once rendering is
completed.

We avoid this problem in `componentDidMount` by deferring this work
until the next tick.

I think we should consider making two changes to Waypoint to help
address this:

1. Extend `React.PureComponent` instead of `React.Component`. In the
   place I saw this problem, we were rendering a waypoint with pure
   props, so the update in this case could be completely avoided by
   using PureComponent.

   ```jsx
   <Waypoint
     onEnter={this.onEnterWaypoint}
     scrollableAncestor={this.props.scrollableAncestor}
   />
   ```

2. Defer `this._handleScroll` in `componentDidUpdate` using
   `onNextTick`, like we do in `componentDidMount`. This will still
   require layout, but it will happen around the same time that the
   browser was going to do layout anyway, avoiding the thrash.

This commit implements the second of these two interventions.

Fixes #263

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -42,9 +42,11 @@ export default class Waypoint extends React.Component {
     }
 
     // this._ref may occasionally not be set at this time. To help ensure that
-    // this works smoothly, we want to delay the initial execution until the
-    // next tick.
-    this.cancelInitialTimeout = onNextTick(() => {
+    // this works smoothly and to avoid layout thrashing, we want to delay the
+    // initial execution until the next tick.
+    this.cancelOnNextTick = onNextTick(() => {
+      this.cancelOnNextTick = null;
+
       // Berofe doing anything, we want to check that this._ref is avaliable in Waypoint
       ensureRefIsUsedByChild(this.props.children, this._ref);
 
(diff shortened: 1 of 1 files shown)


lencioni:
@trotzig @jamesplease PTAL!

PR #278, opened 2018-11-15
Add support for React.forwardRef and React.createRef

Andarist:


trotzig:
Hi @Andarist, and thanks for the contribution! This looks good to me, but I'm a little behind on the latest React features. Can you explain this change a little better? Preferably in the commit message so that future devs can understand why the change was made. More specifically, it would help me if you could explain if this is a breaking change or not (semver major vs semver minor/patch).

Andarist:
I've added a short explanation to the commit message:
> Components wrapped with `React.forwardRef` will now correctly have their refs passed to `react-waypoint` as they don't recognize custom `innerRef` prop, but should know what to do with received `ref` (that is - pass it to underlaying DOM element).

> Also the support for `React.createRef` has been added - which is a new React API, an alternative to callback refs. It returns an object with shape of `{ current: null }`, so the ref can and should be set on the `current` property (it's done by React internally when passingit as ref prop to an element).

I can confirm that this change is backwards compatible. In the future (when doing a major release and dropping support for older React versions) you should dropping the `innerRef` convention entirely, always pass `ref` to the child component and enforce on people usage of `React.forwardRef` if they want a composite component as a child (ideally - you'd also provide a dev only warning when you detect non-forwardRef/DOM child). You can read slightly more about other libraries adopting this strategy [here](https://popmotion.io/blog/20181101-react-pose-4-forwardref/)

trotzig:
Thanks! The only thing I need before merging this PR is an update to the Readme. Will you take care of that?

Andarist:
I'll try to adjust it some time later.

Andarist:
I've adjusted the PR with README change.

trotzig:
Great! Thanks for the quick turnaround.

trotzig:
Released in 8.1.0

PR #284, opened 2018-12-11
[Breaking] Remove prop types in production

realityking:
This will shave a few bytes of every production build. A smart bundler might also notice that `prop-types` is unused and removes that dependency but I wouldn't bet on that.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,3 @@
 {
-  "presets": [["airbnb", { "modules": false }]],
-  "plugins": [["transform-react-remove-prop-types", { mode: 'unsafe-wrap' }]]
+  "presets": [["airbnb", { "modules": false }]]
 }
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
     "babel-loader": "^6.4.0",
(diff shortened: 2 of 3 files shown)


MatthewHerbst:
Why did you choose `loose-envify` over [`babel-plugin-transform-react-remove-prop-types`](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types)?

realityking:
This patch is not actually using `loose-envify`, it's added as a configuration to package.json so consumers of this package don't have to set that up themselves. I'm not the biggest fan of this pattern but it's used by react itself.

Using `babel-plugin-transform-react-remove-prop-types` wouldn't change anything in this regard as `loose-envify` would still be needed when bundling with browserify. I didn't use it because there's only one component in react-waypoint. It's easier to do it manually.

realityking:
After this comment I noticed that `babel-plugin-transform-react-remove-prop-types` is even already used by `react-waypoint` - it just doesn't work 🤦‍♂️ 

I added a commit removing it.

MatthewHerbst:
I feel like I'm missing something. The title of the PR is to remove propTypes from prod, but you're removing the thing that does that rather than properly set it up (or is there some other issue with it)?

realityking:
I think the Babel plugin just severely buggy/not considering all cases.

Since Waypoint is just one component it’s IMO simpler to just manually wrap it.

PR #285, opened 2018-12-11
[Breaking] Require React 15.3 and simplify some code

realityking:
Since I'm proposing a few breaking changes might as well throw this one in. The effect is absolutely tiny but this is as good a time as any to do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ node_js:
   - '8'
   - '6'
 env:
-  - REACT_VERSION=0.14.9
-  - REACT_VERSION=15.0
+  - REACT_VERSION=15.3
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,7 +5,6 @@ node_js:
   - '8'
   - '6'
 env:
-  - REACT_VERSION=0.14.9
   - REACT_VERSION=15.0
```
MatthewHerbst:
```suggestion
  - REACT_VERSION=15.3
```

MatthewHerbst: Can we make this `15.3` to be explicit please?

realityking: Done!

PR #286, opened 2018-12-11
[Breaking] Make Waypoint a named export.

### LINKED ISSUE #238: .default is back in 8.0

realityking:
Fixes #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ yarn add react-waypoint
 ## Usage
 
 ```jsx
-import Waypoint from 'react-waypoint';
+import { Waypoint } from 'react-waypoint';
 
 <Waypoint
   onEnter={this._handleWaypointEnter}
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@ declare class Waypoint extends React.Component<Waypoint.WaypointProps, {}> {
   static invisible: string;
 }
 
-export default Waypoint;
+export Waypoint;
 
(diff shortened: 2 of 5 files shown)


realityking:
I think this is necessary if you wanna ship an ESM. Otherwise users using webpack but not ESM will always need a workaround.

lencioni:
@realityking there are some merge conflicts that need to be resolved here and I'd like to include this in the next release. Can you please rebase this for me?

Also, what do you think about adding the named export in addition to the default export, at least for one version. That would prevent this from being a breaking change and would allow for people to migrate onto the named export more easily.

README.md
```diff
@@ -38,7 +38,7 @@ yarn add react-waypoint
 ## Usage
 
 ```jsx
-import Waypoint from 'react-waypoint';
+import {Waypoint} from 'react-waypoint';
```
lencioni: nit: please add spaces inside these curlies

lencioni:
Would you be down for whipping up a quick codemod that people can use to migrate to the named export?

realityking:
Absolutely but it'd probably have to wait until the weekend. The rest of my week is looking quite busy.

PR #290, opened 2019-01-23
Switch eslint to use airbnb config

realityking:
This PR changed quite a bit scroll down to [this comment](https://github.com/brigade/react-waypoint/pull/290#issuecomment-456981945) to see what's going on.

This PR replaces the unmaintained `eslint-config-brigade` with `eslint-config-airbnb`.

---------
Original message:
This is not the newest version available of `eslint-config-brigade` but otherwise the diff would've become even larger. It's an incremental step to get to `eslint` 4 and `eslint-config-brigade` 7.


lencioni:
Agreed, I think we should switch to the Airbnb eslint config. @realityking would you be interested in modifying this PR to do that instead?

realityking:
I was wondering why that repo hasn't been touched in so long 😆 

AirBnB is the defacto standard in React-land so that seems like a good idea. I quickly threw together a config and it's quite doable. Give me a couple of days and I'll have a PR for you.

realityking:
Done! Let's hope the tests still pass 🤞

realityking:
That should fix it. Not sure how married you are to the `.jsx` extension. Things would be a little easier if it was removed.

PR #291, opened 2019-01-23
Ensure that children is valid only in dev mode

realityking:
This PR does two related things

1. It makes the prop-types validation of `children` more strict. This should help IDEs and linters catch errors earlier.
2. It disables the `React.Children.only` validation in production. This is an error that should be not occur in production since it throws in development and is not the kinda variable that's influence by the user/external data sources.

This will reduce the bundle size a little bit more since minifies can throw away more code.


lencioni:
What do you think about adding the prod check at the callsite? That would allow the function call to be minified out entirely.

realityking:
True. Let me add that.

realityking:
@lencioni Done!

PR #293, opened 2019-01-25
Fix isForwardRef call

### LINKED ISSUE #273: Allow usage of ref on custom components
React 16.3 introduced the concept of [forwarding refs](https://reactjs.org/docs/forwarding-refs.html) as a first-class citizen. We can remove the usage of `innerRef` and just use the regular `ref` prop. We can make this backwards compatible by using `innerRef` if it exists however we shouldn't throw if a custom component doesn't have `innerRef` - `ref` can work just fine.

This library currently breaks with `styled-components@4` due to this issue.

Andarist:
fixes #273

### DIFF
--- a/src/waypoint.jsx
+++ b/src/waypoint.jsx
@@ -295,7 +295,7 @@ export class Waypoint extends React.PureComponent {
       return <span ref={this.refElement} style={{ fontSize: 0 }} />;
     }
 
-    if (isDOMElement(children) || isForwardRef(children.type)) {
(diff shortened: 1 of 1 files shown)


Andarist:
Unfortunately I probably won't have time to do so - im slammed with other PRs and OSS projects. 

Would be good to merge this in as kinda an important bug fix but create an issue for grabs (beginner friendly) for writing this test.

cipater:
I'm running into this issue... any chance this can get merged?

PR #295, opened 2019-02-05
Update README.md

AndrejGajdos:
minor fix in README


AndrejGajdos:
Hmm sorry, when I tried library yesterday, I am sure I needed to use default import

lencioni:
v9.0.0 was just published which should include this change!

PR #296, opened 2019-02-17
Fixed import statement in example

kidunot89:
Simple fix to README example

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ yarn add react-waypoint
 ## Usage
 
 ```jsx
-import { Waypoint } from 'react-waypoint';
+import Waypoint from 'react-waypoint';
 
 <Waypoint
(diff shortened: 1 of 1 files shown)


kidunot89:
https://github.com/brigade/react-waypoint/blob/master/CHANGELOG.md#731. I get an `Waypoint` not found error every time I make an attempt to use the named export `import { Waypoint } from 'react-waypoint';`, but the default export works fine.

MatthewHerbst:
`import Waypoint from 'react-waypoint/es'` is different than `import Waypoint from 'react-waypoint'`. Happy to help you look into your error, but you can clearly see in the code (both the `index.d.ts` file and the `Waypoint.jsx` file) that there is no default export.

lencioni:
v9.0.0 was just published which should include this change!

Issue #297, opened 2019-02-20
Typescript: Declaration or statement expected. 


jadbox:
react-waypoint: 9.0.0
typescript: 3.3.3
```
node_modules/react-waypoint/index.d.ts
Type error: Declaration or statement expected.  TS1128

     8 | }
     9 | 
  > 10 | export Waypoint;
       | ^
    11 | 
    12 | declare namespace Waypoint {
    13 |     interface CallbackArgs {
```

RoberMac:
any workaround?

nucleartux:
I think this issue can be closed.
cc @lencioni / @trotzig

PR #298, opened 2019-02-25
fix export in typings

nucleartux:


ajbouh:
This fixes it for me too – any word on when this might be part of a new release?

MatthewHerbst:
This will fix #297 

cc @lencioni

rosskevin:
Thanks @nucleartux.  Change looks correct and typechecks locally.

This fixes a typing error introduced by @realityking in https://github.com/brigade/react-waypoint/commit/0876d160b9367581f01c1b68b4f0d75da00b00ae#diff-b52768974e6bc0faccb7d4b75b162c99

Just a note, in-repo types cause big problems for downstream ts users - our only solution is to lock it back to an old version.  Since this is a needed bug fix for ts users, can we get a merge/patch release soon?

PR #299, opened 2019-02-28
Test on Node.js using Jest

realityking:
This PR moves those tests that are not browser specific to Jest.

This also allowed cleaning up the `getWindow()` workaround to simulate a non-browser environment in a browser.

Technically this is semver-major as it removes an exported method. Since the method is not documented (not even as part of the TypeScript definition), I think it's worth the risk.


src/waypoint.jsx
```diff
@@ -370,11 +372,5 @@ Waypoint.above = ABOVE;
 Waypoint.below = BELOW;
 Waypoint.inside = INSIDE;
 Waypoint.invisible = INVISIBLE;
-Waypoint.getWindow = () => {
```
lencioni: It looks like this was added by 4e22e27e9290bba61f4c133509cc8e2611e72079 but I'm not really sure why. @trotzig do you happen to know?

realityking: I believe it was only added to allow this mock: 

https://github.com/brigade/react-waypoint/commit/4e22e27e9290bba61f4c133509cc8e2611e72079#diff-27fadc9a7e55db698d635052018a532aR424

lencioni: Cool, yeah I think this is okay to remove!

realityking:
I've updated this PR to consider more recent changes. I think it's ready to merge.

It currently includes the changes from #332 so it could be tested. I'll remove these once that PR has landed.

realityking:
It took a while but this PR is finally in a state where I think it's ready to go.

Note that the final commit could be considered a breaking change. I think it's very low risk since the method was never documented but let me know if you want me to replace it with a deprecation message.

realityking:
@trotzig I have a couple of other things in mind. Can you wait ~1 week for the next release?

trotzig:
Will do! Let me know when you're ready.

PR #300, opened 2019-02-28
Remove the custom warning for multiple children to support React's StrictMode.

### LINKED ISSUE #294: StrictMode warnings
Couldn't find an issue about this, but StrictMode results in two warnings:
screen shot 2019-01-26 at 18 41 30

The two cases are:
- https://github.com/brigade/react-waypoint/blob/master/src/waypoint.jsx#L34
- https://github.com/brigade/react-waypoint/blob/master/src/waypoint.jsx#L79

Which are just running checks on props.children, so it seems like it would be trivial to fix by using `componendDidMount` and `static getDerivedStateFromProps` respectively.

Let me know if there's something I'm missing, but I'd be happy to submit a PR for it if that's desired.

realityking:
Fixes #294

I tried for a couple of hours to make it work using `react-lifecycles-compat` but the use-case just isn't well covered by the new React API.

I think it's ok to remove this error because:

- #291 added the correct validation to the `propTypes`. React will warn if it's used incorrectly.
- React 16.3 throws a fatal error anyway when it's used like this (though with a not self explanatory error message)

WDYT?


src/waypoint.jsx
```diff
@@ -10,7 +10,6 @@ import {
   INVISIBLE, INSIDE, BELOW, ABOVE,
 } from './constants';
 import debugLog from './debugLog';
-import ensureChildrenIsValid from './ensureChildrenIsValid';
```
lencioni: If this file is unused, can we please also remove it in this PR?

realityking: Ups, that slipped through the cracks. Fixed now.

PR #326, opened 2020-06-18
Add support for scrollable ancestor with overflow: overlay

mohammadalipak:
Scrollable ancestors can also have `overflow: overlay;`. Currently, react-waypoint does not support that case. This PR adds that support.


trotzig:
The test failures aren't caused by your change, so don't worry about those. 

Are you using this property yourself? It seems like this is a webkit-only property, I hadn't seen it before.

trotzig:
Released in v9.0.3

PR #332, opened 2020-09-27
Fix testing on TravisCI and upgrade node versions

realityking:


realityking:
Any chance this could be merged to unblock CI for other PRs?

trotzig:
Sorry about the delay here -- this looks good. It would have been good if there was a lengthier description in the commit message or in the PR. There are some changes that on a surface level don't seem obvious (mainly thinking about the difference in how we now specify Node versions to test). That being said, I'm not going to let this unblock other changes.

realityking:
Sorry about that. Let me elaborate here so it's documented for prosperity:

- `dist: bionic` moves the build process to a newer Ubuntu version.
- `services: xvfb`, this is the main change to get this working again. It's due to a breaking change from [TravisCI](https://benlimmer.com/2019/01/14/travis-ci-xvfb/).
- The change to react versions is mainly to avoid having to regularly update the config with new node versions. Since `react-waypoint` is mainly a browser library and there's very little risk for incompatibilities between Node.js versions it just tests on the most current stable version and the most recent LTS version.
- Lastly it removes the manual installation of npm 6.4.1 since all node versions tested on have more recent versions

Issue #322, opened 2020-03-23
Play video only when element is fully visible in the ui


svankireddy-cartera:
Currently video is played even when video is partially viewable.
I want to only play video when video is fully viewable. 

Is there any callback function I can use for this?

```
import React from 'react';
import { Waypoint } from 'react-waypoint';

class  AutomaticVideoPlayer extends React.Component {
  constructor(props) {
    super(props);
    this.videoElementRef = React.createRef();
  }

  handleEnterViewport() {
    const videoRef = this.videoElementRef;
    console.log(videoRef);
    videoRef && videoRef.play();
  };

  handleExitViewport() {
    const videoRef = this.videoElementRef;

    videoRef && videoRef.pause();
  };

  render() {
    return (
      <Waypoint onEnter={this.handleEnterViewport} onLeave={this.handleExitViewport}>
        <video ref={this.videoElementRef} playsInline autoPlay loop muted {...this.props} />
      </Waypoint>
    );
  }
}

export default AutomaticVideoPlayer;
```

trotzig:
Hi @svankireddy-cartera! I would probably look into using `IntersectionObserver` for your use-case. Since you're using React, you could look at https://www.npmjs.com/package/react-intersection-observer

svankireddy-cartera:
@trotzig  thank you for suggestion

PR #334, opened 2020-10-28
Disable TravisCI's npm cache

realityking:
It seems to conflicts with the diferent npm versions used in the matrix.

This is because of my change in #322 that always tests on the lasted stable Node.js version and the release of Node.js 7 that bundles npm 7. I think it's good these gets covered in testing but the cache seems to be a problem.


trotzig:
Thanks!

PR #335, opened 2020-11-09
feature props delay onEnter

aswara:


trotzig:
Hi @aswara, and thanks for the PR! In order to get things like this merged, you'll need to provide a thorough description of the feature you're adding. Some things to cover: 
- What is the feature? 
- Why are you adding the feature? 
- Why is the feature needed in this library? I.e. explain why you can't do this yourself without too much hassle.

PR #336, opened 2020-11-12
Support React 17

realityking:
This PR adds testing for React 17 and relaxes the peer dependency constraint on `react` to not warn for React 17

It also updates `react-is` to version 17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,6 @@ cache:
 env:
   - REACT_VERSION=15.3
   - REACT_VERSION=16.0
+  - REACT_VERSION=17.0
 before_script:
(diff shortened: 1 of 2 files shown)


realityking:
@trotzig This is one that should definitely make it I to the next release

lencioni (approved):
Thanks!

plotka:
@lencioni  When is the next release that will include support for React 17?

mediabeastnz:
👀

tommoor:
Howdy, seems like this was merged 6months ago but there still hasn't been a release since then. React 17 has been out quite a while now!

lencioni:
This has been published with v10.0.0!

PR #339, opened 2020-11-17
Use .mjs for module build

realityking:
This will improve compatability with Node.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "9.0.3",
   "description": "A React component to execute a function whenever you scroll to an element.",
   "main": "cjs/index.js",
-  "module": "es/index.js",
+  "module": "es/index.mjs",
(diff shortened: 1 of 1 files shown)


trotzig:
We had this a (long) while back but had to revert because it broke things for a lot of people. It might be worth reading up on some discussions there: 

Introduced in 
https://github.com/civiccc/react-waypoint/pull/220

Reverted in 
https://github.com/civiccc/react-waypoint/pull/223

There's been a lot of water under the bridge since then however, and it might be just fine to make this change today.

realityking:
The most relevant discussion seem to be in #234 and #221 

I _think_ this might've been fixed by #286 but I'm also not confident so I'm gonna close this one.

Until `react-waypoint` actually adds an export map to run modules in Node it doesn't give any benefit anyway.

PR #341, opened 2021-01-26
Use external @babel/runtime helpers

realityking:
Fixes #302

It does complicate the rollup config quite a bit but this is needed to correctly generate both the ESM and CJS versions.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,4 +1,4 @@
-import babel from '@rollup/plugin-babel';
+import babel, { getBabelOutputPlugin } from '@rollup/plugin-babel';
 import pkg from './package.json';
 
 export default [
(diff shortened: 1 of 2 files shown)


trotzig:
Do you know if this increases the size of the built javascript file?

realityking:
Do you mean the bundled file including the Babel helpers? It will probably depend on how things are build. If other libraries or the application itself uses `@babel/runtime` it'll be smaller, if the build pipeline is set-up for ESM and concatenating modules (web pack 4 or 5) it should be at worse the same. If it's set-up for CommonJS and there's no deduplication it will be a bit bigger because every module is wrapped.

I'll try to get some data on this :)

realityking:
To get an idea of the size impact, I created a couple of different builds of the performance tests. This is using Webpack 4, I'd expect better results with Webpack 5 or Rollup.

- Current setup (build of src): 144K
- ESM build with external helpers: 144K
- CJS build with external helper: 148K
- ESM build without external helpers: 144K
- CJS build without external helper: 148K

Conclusion: CJS sucks but no impact on externalising these helpers :)

realityking:
@trotzig I think a release needs to be done to address #347. It'd be great if you could take a look if you're comfortable including this in that release, if not, please go ahead without this one.

lencioni (approved):

rollup.config.js
```diff
@@ -9,12 +9,34 @@ export default [
...
+            modules: false,
+            runtimeVersion: '7.12.5',
+            runtimeHelpersUseESModules: true,
+          }]],
+        })],
+      },
     ],
     plugins: [
       babel({
+        babelrc: false,
```
lencioni, line 37: I think this means that we probably want to remove the production section of the .babelrc. I can do that separately.

rollup.config.js
```diff
@@ -9,12 +9,34 @@ export default [
...
         babelHelpers: 'bundled',
```
lencioni, line 38: Although don't we want to change this to `runtime` instead of using the getBabelOutputPlugin method? I'll try out an alternative approach now.

lencioni: https://github.com/civiccc/react-waypoint/pull/353

lencioni:
I did this slightly differently here https://github.com/civiccc/react-waypoint/pull/353

Thanks for opening this PR!

realityking:
That approach is entirely fair as well. :) The different is really one of personal preference as I don't like complicated `.babelrc` files 😆 

Thanks for getting this merged!

PR #342, opened 2021-01-26
Upgrade test related dependencies & adapt tests

realityking:
This one was interesting as quite a few things broke when upgrading stuff.

- Karma to version 6 worked as-is.
- `jasmine-core` to 2.99 was a relatively simple change. I believe the test has never worked, jasmine-core 2.7 just exposed that
- On that note, I couldn't figure out how to get up to `jasmine-core` 3
- Webpack 3 just worked but Webpack 4 required some changes as Webpack < 4 didn't correctly execute code in strict mode.

I think most of these changes are no brainers except perhaps the upgrade to webpack. I think it's valuable to test with older webpack versions to assure compatibility. However webpack 4 is now almost 3 years old. I think it's fine to ignore these older versions. Especially since it uncovered tests weren't actually executed in strict mode, potentially hiding bugs.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -27,12 +27,18 @@ module.exports = (config) => {
     },
 
     webpack: {
+      mode: 'development',
       module: {
-        loaders: [
+        rules: [
           {
             test: /\.jsx?$/,
-            loaders: ['babel-loader?cacheDirectory=true'],
             exclude: /node_modules/,
+            use: {
+              loader: 'babel-loader',
+              options: {
+                cacheDirectory: true,
+              },
+            },
           },
         ],
       },
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
     "@rollup/plugin-babel": "^5.2.1",
-    "@types/react": "^15.0.21",
+    "@types/react": "^16.14.5",
     "babel-loader": "^8.0.0",
     "babel-preset-airbnb": "^5.0.0",
     "eslint": "^7.12.0",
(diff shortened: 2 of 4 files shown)


trotzig:
Thanks -- looks good to me!

The test failure in Node v15 is due to a change in node how peer dependencies are handled. I think this could be fixed by adding `react-dom` as a peer dependency (which I think should already have been made). You could also use `npm install --legacy-peer-deps` as a workaround (although I'm not sure how to configure that for Travis).

realityking:
Yeah npm 7 is an pain in the ass. I wonder how tests passed on master - none of the changes in either PR really touched the React dependency. I'll have a look what can be done.

What's your take on the webpack question? Good to go with webpack 4?

trotzig:
I think what happened is that the node version named "node" (!) on Travis got updated from 14 to 15 at some point. 

Yep, webpack 4 is fine. 👍

realityking:
I thought so too but the last passing build already run with nom 7.0.8: https://travis-ci.org/github/civiccc/react-waypoint/jobs/743936673#L196

realityking:
Quickly tested locally

- node v15.4.0 (npm v7.0.15): works ✅
- node v15.5.1 (npm v7.3.0): breaks 🛑

Deviously, it works when the `node_modules` folder already exists but it breaks when doing a clean install. I think this might be an npm bug.

realityking:
The issue has been already been reported to npm: https://github.com/npm/cli/issues/2514

I see two possible ways out of this until this is fixed:

1. Force `--legacy-peer-deps` for npm
2. Pin the node version to 15.4 and thus npm to 7.0

Which one do you prefer?

realityking:
@trotzig The npm bug has been fixed and CI passes. This is good to go from my perspective.

realityking:
@trotzig It'd be great if you could have a look at this one :)

PR #344, opened 2021-02-04
Fix fire `onEnter` callback anyway when parent isn't display

### LINKED ISSUE #328: When parent is hidden, it will fire onEnter callback anyway
If I place `<Waypoint />` in a hidden parent container, it will fire `onEnter` callback immediately as soon as it mounted, even if it's not in viewport. Which seems is a bug?

Here is the reproduce demo: https://codesandbox.io/s/react-waypoint-bug-nr0j4

xyy94813:
Fix #328


xyy94813:
Is there any maintainer?
Give me some feedback plz.

ferleonardo (approved):

PR #349, opened 2021-03-11
docs: update "bottomOffset" description

abramenal:
### Current state

2021-03-11 4 50 53 PM

### Desired state

The similarity (or even "specularity") of "topOffset" and "bottomOffset" is kinda obvious, but still does make sense to describe both of them properly (similar to `onEnter` vs `onLeave`)


lencioni:
Thanks!

PR #362, opened 2022-06-02
REACT18 TS - add support for children props

ederelk:
REACT18 TS - Children are no longer  implicit property of a Component.
Children has to now be specified explicitly on the props declaration.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -110,5 +110,10 @@ declare namespace Waypoint {
          * things down significantly, so it should only be used during development.
          */
         debug?: boolean;
+      
+        /**
+         * Since React 18 Children are no longer implied, therefore we specify them here
+         */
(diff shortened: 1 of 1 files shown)


sanderkooger:
@ederelk Thanks for making the PR. Hope they merge soon!

lencioni (approved):
Thanks!

lencioni:
Unfortunately, it seems that I have lost the ability to publish this package somehow. You might want to consider either using patch-package (https://www.npmjs.com/package/patch-package) or switching to a different thing, like IntersectionObserver (maybe via react-intersection-observer https://www.npmjs.com/package/react-intersection-observer).

trotzig:
I managed to push v10.3.0 🎉

ederelk:
Thank you very much for merging the update

